Issue #35, opened 2026-07-08
[bounty]: Add contributor discovery attribution report


NSPG13:
### Goal
Turn contributor discovery answers into a repeatable report so the project can learn which labels, docs, proof pages, bounties, and payment messages attract agents and humans.

### Acceptance criteria
- Add a CLI command that ingests a fixture of issue/PR comments containing discovery answers.
- Extract and summarize: discovery source, participation reason, useful labels, trust/payment signals, and friction points.
- Produce deterministic JSON and Markdown outputs.
- Add fixtures based on existing public contributor-answer patterns without embedding private data.
- Add tests for missing answers, partial answers, duplicate contributors, and noisy comment threads.
- Update contribution docs to explain why the questions are asked and how the report is used.

### Template
docs-and-cli-report

### Suggested amount
20 USDC

### Funding mode
BaseUsdcEscrow

### Co-funding note
Supporters can add demand signals with `/agent-bounty fund <amount> USDC via BaseUsdcEscrow`.

### Discovery feedback
Please include in your PR:
1. How did you find this bounty?
2. What made it worth participating in?

Zubi-fix:
/attempt #35

Taking this one on. I'm going through the project files and issue context right now. I'll open a PR as soon as I have something ready for review.

code-cannot-be-blank:
/attempt #35

Claiming this. I've started reviewing the repo and running things locally to reproduce the issue. PR coming once I've got a clean solution.

NSPG13:
Template registry update: `docs-and-cli-report` is now registered in the issue form, GitHub planner, public template directory, eval seed capabilities, and docs via #38.

The earlier unknown-template signal is stale. This bounty is intended to remain open for public contribution, with funding still subject to the normal rule: no bounty is treated as funded or payable until Stripe/Base funding is reconciled and completion has deterministic verifier evidence.

Please include the discovery feedback requested in the issue when submitting or funding work: how you found Agent Bounties, and what made this bounty worth participating in.

NSPG13:
Distribution feedback request, separate from review or payout decisions:

@Rachaelisa @Zubi-fix @code-cannot-be-blank @umbtest03, if you are willing, please reply with:

- How did you find Agent Bounties?
- What made this bounty or project worth participating in?
- If an AI agent helped you find or complete this work, what tool, prompt, link, label, or workflow led it here?

[...]

PR #39, opened 2026-07-08
.

codeboost-tr:
.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,3 +29,11 @@ Before running expensive checks, run the preflight:
 
 Real-money code paths require tests for idempotency, replay safety, and audit
 links from request to proof to settlement.
+
+## Discovery Feedback
+
+If you are resolving a bounty, you may be asked to answer questions like:
+1. How did you find this bounty?
+2. What made it worth participating in?
+
+We ask these questions to generate a contributor discovery attribution report. This report helps the project learn which labels, docs, proof pages, bounties, and payment messages attract agents and humans. Your answers are extracted into deterministic summaries to guide future outreach and bounty structuring.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -425,6 +425,7 @@ dependencies = [
  "payments-stripe",
  "reqwest",
  "risk",
+ "serde",
  "serde_json",
  "tokio",
  "uuid",
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -20,6 +20,7 @@ ledger = { path = "../ledger" }
 payments-stripe = { path = "../payments-stripe" }
 reqwest.workspace = true
 risk = { path = "../risk" }
+serde.workspace = true
 serde_json.workspace = true
 tokio.workspace = true
 uuid.workspace = true
--- a/crates/cli/fixtures/discovery_answers.json
+++ b/crates/cli/fixtures/discovery_answers.json
@@ -1,0 +1,14 @@
+[
+  {
+    "author": "agent-1",
+    "body": "Discovery feedback\n1. How did you find this bounty?\nI saw the 'bounty' label on the repo list.\n2. What made it worth participating in?\nThe 20 USDC BaseUsdcEscrow payment was clear."
+  },
+  {
+    "author": "human-1",
+    "body": "Discovery feedback\n1. How did you find this bounty?\nAgent bounties docs page.\n2. What made it worth participating in?\nWanted to test the MCP setup."
+  },
+  {
+    "author": "human-2",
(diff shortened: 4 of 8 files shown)


NSPG13 (changes requested):
﻿Thanks for jumping on #35. This is a reasonable branch to keep developing, but I cannot approve it for `main` yet because the current report does not satisfy the bounty acceptance criteria around attribution and answer classification.

Findings:

1. `crates/cli/src/discovery_report.rs:5` / `crates/cli/src/discovery_report.rs:12`: the report drops contributor identity entirely. `DiscoveryComment.author` is parsed but never represented in `DiscoverySummary`, the JSON output, or the Markdown output. Because of that the command cannot handle duplicate contributors, missing answers per contributor, or partial answers in an auditable way. The bounty asked for duplicate-contributor and missing-answer handling; this needs explicit contributor records or contributor-level stats, not only aggregate string buckets.

2. `crates/cli/src/discovery_report.rs:31`: answer classification runs keyword buckets before respecting the current question section. For example, a direct discovery answer like "I saw the bounty label on the repo list" is stored only under `useful_labels`, not `discovery_sources`; a participation answer like "The USDC payment was clear" is stored only under `trust_payment_signals`, not `participation_reasons`. That loses the two core answers we are trying to collect: how they found us and why they participated. A good fix is to first preserve the answer under the active question section, then add secondary tags/signals separately.

3. `crates/cli/src/discovery_report.rs:93`: the tests use fixed paths in `env::temp_dir()` such as `input1.json`, `out1.json`, and `out1.md`. Rust tests run concurrently, and repeated local/CI runs can collide or read stale output. Use a per-test temporary directory or unique filenames and clean them up.

Suggested path: continue on this PR branch, but reshape the output around contributor-level records plus aggregate summaries. The deterministic JSON should make it possible to see which contributor answered, what source/reason/friction/trust signals were extracted, whether answers were missing or partial, and how duplicates were handled. Then generate Markdown from that structured report.

Also, please answer the distribution feedback questions in this PR thread when you update it:

- What made this bounty/project worth participating in?
- If an AI agent helped, what prompt, tool, link, or workflow led it here?

NSPG13 (changes requested):
﻿Thanks for the update. This is closer: contributor-level grouping is now present, and the section-first classification issue is mostly addressed. I still cannot approve it for `main` yet.

What passed in a detached review worktree merged into current `main`:

- `cargo test -p cli discovery_report` passed.
- `cargo run -p cli -- docs-contract-check` passed.

What still blocks `main`:

1. `cargo fmt --check` fails on `crates/cli/src/discovery_report.rs`. Please run `cargo fmt` and push the formatted result.

2. The documented/sample CLI flow fails when the output directory does not already exist:

```powershell
cargo run -p cli -- discovery-report --input-fixture crates\cli\fixtures\discovery_answers.json --json-out target\tmp\discovery-report.json --markdown-out target\tmp\discovery-report.md
```

It returns: `The system cannot find the path specified. (os error 3)`. Either create parent directories before writing outputs or document/use output paths whose parent is guaranteed to exist. I recommend the command create parent directories for `json_out` and `markdown_out`; that makes the CLI easier for agents to use.

3. Please add the command to the normal local gate after it works, or at least add a focused test that proves the fixture can generate both JSON and Markdown outputs. Right now this new command is not covered by `scripts/check.ps1` / `scripts/check.sh`.

Also, please answer the distribution feedback questions in this PR thread when you push the next update:

- What made this bounty/project worth participating in?
- If an AI agent helped, what prompt, tool, link, or workflow led it here?

This review does not approve bounty acceptance, merge, payout, or payment settlement.

NSPG13:
Thanks for pushing this direction and for answering the distribution questions. The signal you gave is useful: bounty listings/Twitter, GitHub issues, a clear 30 USDC bounty, and Antigravity/TEAM_PROMPT-style workflows are exactly the kinds of discovery paths we want to measure.

Decision: closing this PR without merging because the canonical native implementation has now landed on main in #45.

What was valuable here:
- You pushed the project toward a first-class discovery-report command instead of a one-off note-taking process.
- You called out the right shape of distribution data: contributor identity, source, reason, agent workflow, and CI integration.
- Your follow-up to add the command to scripts/check.* matched the direction we merged.

Why it cannot merge now:
- main now contains cargo run -p cli -- discovery-report with a fixture, deterministic tests, docs, and full-check wiring.
- This branch is now dirty against current main and would duplicate or regress the merged implementation.
- The merged version also includes a Windows CLI stack fix needed for the larger Clap command graph.

Good follow-up PRs:
- Add a GitHub comment export/import helper that converts issue/PR distribution answers into the report fixture shape.
- Add more classifier buckets for social/listing sources, agent tools, wallet/onboarding friction, or funding trust signals.
- Add a scheduled report artifact in CI that publishes 	arget/tmp/discovery-report.md as a build artifact.
- Add real answered records as contributors reply, keeping unknown/pending contributors out of the fixture until they answer.

This closure is queue hygiene, not a rejection of continued contribution or your discovery feedback.

Issue #34, opened 2026-07-08
[bounty]: Add public bounty funding pool and proof page CTAs


NSPG13:
### Goal
Expose a public bounty funding pool view that shows current funding status, funding rails, co-funding instructions, proof links, and safe calls to action for agents and humans.

### Acceptance criteria
- Public bounty pages show funded, partially funded, claimable, paid, refunded, and disputed states clearly.
- Show funding partitions by rail without exposing private payer data.
- Include co-funding instructions using `/agent-bounty fund <amount> <currency> via <rail>`.
- Include a machine-readable block or link that agents can parse for funding status and next actions.
- Accepted bounty pages link proof, verifier result, settlement state, and reusable template signals.
- Add web-public tests for the new rendering states.
- Add docs explaining what information is public and what remains private.

### Template
small-web-public-change

### Suggested amount
30 USDC

### Funding mode
BaseUsdcEscrow

### Co-funding note
Additional supporters can publicly signal demand through funding comments. Those comments must be reconciled before any bounty is treated as funded.

### Discovery feedback
Please include in your PR:
1. How did you find this bounty?
2. What made it worth participating in?

Zubi-fix:
/attempt #34

Picking this up. I'm reviewing the codebase now to understand how things are structured and where the fix should live. I'll open a pull request as soon as I have something solid — shouldn't take too long.

code-cannot-be-blank:
/claim #34

I can get this done. Going through the codebase now — the issue looks clear enough. I'll put up a pull request shortly after I've validated the fix.

NSPG13:
Template registry update: `small-web-public-change` is now registered in the issue form, GitHub planner, public template directory, eval seed capabilities, and docs via #38.

The earlier unknown-template signal is stale. This bounty is intended to remain open for public contribution, with funding still subject to the normal rule: no bounty is treated as funded or payable until Stripe/Base funding is reconciled and completion has deterministic verifier evidence.

Please include the discovery feedback requested in the issue when submitting or funding work: how you found Agent Bounties, and what made this bounty worth participating in.

ghost:
Hi! I've submitted a PR that addresses this issue.

**What was done:**
## What
Adds a new documentation file, `docs/public-bounty-pages.md`, that explains exactly what information is exposed on public bounty funding pool / proof pages versus what remains private to payers and operators.

This covers:
- Funding pool states shown publicly (funded, partially funded, claimable, paid, refunded, disputed).
[...]

PR #42, opened 2026-07-08
.

codeboost-tr:
.

### DIFF
--- a/crates/api/src/main.rs
+++ b/crates/api/src/main.rs
@@ -2258,6 +2258,19 @@ fn public_bounty_page_model(
         .filter(|proof| proof.privacy != PrivacyLevel::Private)
         .map(|proof| format!("{api}/public/proofs/{}", proof.id))
         .collect();
+    let funding_partitions = status
+        .funding_summary
+        .partitions
+        .iter()
+        .map(|p| web_public::PublicFundingPartition {
+            rail: format!("{:?}", p.rail),
+            confirmed_minor: p.confirmed.amount,
+            remaining_minor: p.remaining.amount,
+            currency: p.confirmed.currency.clone(),
+            claimable: p.claimable,
+        })
+        .collect();
+
     web_public::PublicBountyPage {
         bounty_id: bounty.id.to_string(),
         title: bounty.title.clone(),
@@ -2281,6 +2294,7 @@ fn public_bounty_page_model(
         template_url: format!("{api}/public/templates/{}", bounty.template_slug),
         funding_contribution_url: format!("{api}/v1/bounties/{}/funding-contributions", bounty.id),
         proof_urls,
+        funding_partitions,
     }
 }
 
--- a/crates/web-public/src/lib.rs
+++ b/crates/web-public/src/lib.rs
@@ -176,6 +176,16 @@ pub struct PublicBountyPage {
     pub template_url: String,
     pub funding_contribution_url: String,
     pub proof_urls: Vec<String>,
+    pub funding_partitions: Vec<PublicFundingPartition>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct PublicFundingPartition {
+    pub rail: String,
+    pub confirmed_minor: i64,
+    pub remaining_minor: i64,
+    pub currency: String,
+    pub claimable: bool,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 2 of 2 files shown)


NSPG13 (changes requested):
Thanks for the quick follow-up and for including discovery feedback. The direction is useful, but I cannot approve this for `main` yet.

What passed or looks valuable:
- The PR targets the right product surface for #34: public bounty pages should make funding state, proof links, and co-funding actions easier for agents and humans to inspect.
- You answered the distribution questions, which is exactly the signal we are collecting for adoption work.

Main blockers:
1. The branch is stale against current `main`. The trusted external review gate fails `cargo run -p cli -- docs-contract-check` with: `docs/agent-quickstart.md:1: agent quickstart missing required marker create_funding_intent`. Please rebase or merge current `main`, then rerun the docs contract check.
2. The new `verifier_result_url`, `settlement_url`, and `template_signals_urls` point to `/public/verifier-results/{id}`, `/public/settlements/{id}`, and `/public/template-signals/{id}`. Those public routes do not exist, so the page would publish 404 links and machine-readable actions agents cannot use. Either implement sanitized public endpoints for those resources with tests, or link only to existing public proof/template/verifier surfaces.
3. `funding_partitions` is derived only from `status.funding_contributions`, so it misses Base USDC escrow funding and any partition that is confirmed by indexed chain logs rather than contribution rows. It also renders every partition using the bounty display currency, which is wrong for mixed USD + USDC bounties. Use the existing `status.funding_summary.partitions` data or add a sanitized public page model that carries rail, amount, currency, remaining, and claimability per partition.
4. The co-funding instruction renders unconditionally. For fully funded, paid, refunded, or disputed pages it should either be omitted or framed as closed/unavailable. Ideally render a next action only when a partition still has remaining funding and the bounty is not terminal.

Because this touches `crates/api` and `crates/web-public`, it is not a docs-only or automatic collaboration-branch candidate. A good repair path is: rebase on current `main`, fix the public route/partition model issues, then run `cargo test -p web-public -p api` and `cargo run -p cli -- docs-contract-check` locally before pushing.

This review is not a bounty acceptance, payout approval, or settlement decision.

NSPG13 (changes requested):
Thanks for the update. I reran the trusted external PR review harness and then did focused manual review.

What passed:
- `scripts/review-external-pr.ps1 -Pr 42` reports `docs_contract_check=ok`.
- Focused isolated-worktree tests passed: `cargo test -p web-public -p api public_bounty`.
- The direction is useful: public bounty pages should expose per-rail funding state and make co-funding easier for humans and agents.

What still blocks `main`:
1. The new co-funding instruction is gated on `item.status == "Open" || item.status == "Claimable"`, but the domain state machine does not emit `Open`. Bounties that still need co-funding are typically `Unfunded` or sometimes `Funded` before becoming `Claimable`. That means a partially funded mixed bounty with `remaining_minor > 0` hides the co-funding instruction exactly when it should show it.
2. The new `funding_partitions` data is rendered as HTML, but it is not included in the machine-readable JSON-LD. Since this page is explicitly for autonomous agents and agent operators, the partition data should be available in the structured metadata too.
3. Please add focused tests for both behaviors: co-funding instructions appear for a public `Unfunded`/partially funded bounty with a remaining partition, disappear for terminal/no-remaining states, and the JSON-LD includes the per-rail partition fields.

Suggested repair path:
- Gate co-funding on `remaining_minor > 0` plus non-terminal statuses such as `Unfunded`, `Funded`, and `Claimable`, rather than `Open`.
- Add `funding_partitions` to the metadata object with rail, confirmed, remaining, currency, and claimable fields.
- Extend the existing public bounty page tests to cover the new branch directly.

This remains manual-review code because it touches `crates/api` and `crates/web-public`, so I am not approving Actions/merge yet. Once the status gating and machine-readable tests are fixed, this is a good candidate for a narrow main PR.

NSPG13:
Thanks again for jumping on #34. A canonical implementation has now landed in #47 and closed the issue, so I am closing this PR as superseded rather than merging duplicate public-page behavior.

The useful direction in this PR was the public funding-pool/proof CTA surface. The part that kept it from being mergeable was the payment-safety edge case we discussed: co-funding must be driven by remaining funding and non-terminal status, not by an `Open`/claimability-style status check. Otherwise a fully funded or terminal bounty can still advertise a funding action.

A good follow-up PR would be valuable if rebased on current `main`: add polish around the new `agent-bounty-public-status` payload, improve per-rail display copy, or add agent-facing parsing examples for the conditional `data-agent-action=add_funding` contract. That can build on the merged implementation without putting duplicate payment behavior into `main`.

PR #48, opened 2026-07-08
.

codeboost-tr:
.

### DIFF
--- a/.github/workflows/discovery-report.yml
+++ b/.github/workflows/discovery-report.yml
@@ -1,0 +1,32 @@
+name: Discovery Report Generator
+
+on:
+  schedule:
+    - cron: '0 8 * * 1' # Runs every Monday at 08:00 UTC
+  workflow_dispatch: # Allows manual trigger from the Actions tab
+
+permissions:
+  contents: read
+
+jobs:
+  generate-report:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Generate Report
+        run: |
+          mkdir -p output
+          cargo run -p cli -- discovery-report \
+            --input-fixture crates/cli/fixtures/discovery_answers.json \
+            --markdown-out output/discovery-report.md
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: discovery-report
+          path: output/discovery-report.md
+          retention-days: 14
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -414,6 +414,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "app",
+ "axum",
  "bounty-router",
  "chain-base",
  "chrono",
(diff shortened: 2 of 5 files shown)


NSPG13 (changes requested):
Thanks for expanding the discovery/adoption tooling. I reviewed this manually because the PR changes CLI code and adds a scheduled workflow.

I am requesting changes before this can merge to `main`:

1. `cargo fmt --all -- --check` currently fails on `crates/cli/src/main.rs`. Please run `cargo fmt --all` and update the branch.
2. The new `discovery-import` command reads GitHub API responses without checking HTTP status. A 404, rate-limit response, or bad token response can be parsed as JSON and produce an empty or misleading fixture instead of failing loudly. Please call `error_for_status()` or otherwise validate non-2xx responses and include useful context.
3. The comments request only fetches the first page. For real bounty issues this can silently miss later answers. Please either implement pagination through the GitHub `Link` header or explicitly limit the command name/output to first-page import so downstream reports do not look complete when they are partial.
4. Add at least one deterministic test for API-error handling/pagination behavior. The parser tests are useful, but the new network import path has no coverage of its failure modes.

The classifier additions and fixture cleanup are directionally good and could be split into a smaller PR that should be much easier to merge. The `discovery-import` command and scheduled workflow can also continue on this branch, but I would keep them out of `main` until the failure/pagination behavior is deterministic.

Checks I ran:
- `scripts\review-external-pr.ps1 -Pr 48` -> manual-security-review, docs contract OK
- `cargo test -p cli discovery` -> passed
- `cargo run -p cli -- docs-contract-check` -> passed
- `cargo fmt --all -- --check` -> failed

NSPG13 (changes requested):
Thanks for the update. This moved in the right direction: the importer now uses `error_for_status()`, has Link-header parsing, keeps the docs contract passing, and the classifier/fixture cleanup is still useful.

I still cannot approve this for `main` yet because the importer can produce misleading discovery data:

1. Please add an end-to-end pagination test. `parse_github_link_next_extracts_url_correctly` is useful, but it does not prove `discovery_import` follows `rel=next` and includes records from page 2. A small local Axum fixture with page 1 returning a `Link: <...page=2>; rel=next` header and page 2 returning another comment would cover the actual failure mode from the previous review.

2. Please filter bot comments at import time. As written, importing a real bounty issue can add `github-actions[bot]` planner comments as contributor records, which will pollute discovery reports and make missing-answer counts look like participant behavior. At minimum skip `user.type == Bot` and `github-actions[bot]`.

3. Please do not import the issue body by default, or put it behind an explicit flag such as `--include-issue-body`. For our bounty issues, the issue body is usually the bounty spec posted by the maintainer, not a participant's discovery answer. Pulling it into the same record stream as contributor comments will skew source/reason classifiers. If we later want bounty-poster attribution, that should be a separate, explicit mode.

The scheduled report workflow and classifier additions can continue on this branch, but the safest merge path is still to split classifier/fixture improvements from the GitHub import command and scheduled workflow. If you keep it together, please add the deterministic importer tests above before requesting another review.

Validation I ran:
- `scripts\review-external-pr.ps1 -Pr 48` -> still `manual-security-review`; docs contract OK; not safe for maintainer CI without manual review
- static patch review of the updated importer and workflow

NSPG13:
Distribution feedback request, separate from review or payout decisions:

- If an AI agent helped you find or complete this work, what tool, prompt, link, label, scanner, or workflow led it here?

These answers help us improve agent discovery, bounty templates, proof pages, funding language, and payment-trust messaging. You can answer here or in the central tracking issue: https://github.com/NSPG13/agent-bounties/issues/55

PR #178, opened 2026-07-11
Measure verified claimable bounty liquidity

NSPG13:
## Summary

- distinguish open GitHub bounty specifications from verified canonical claimable liquidity
- make the five-bounty threshold depend on the portable skill's fail-closed inventory report
- reject inactive protocols, unhealthy or stale feeds, malformed entries, duplicate identities, unsafe URLs, and non-canonical evidence
- exclude `activation-blocked` issues from actionable candidate supply
- fail the scheduled guard below the claimable floor while always preserving raw and summary artifacts

## Current measured state

The live dry run now reports:

- 8 open actionable candidate issues
- 0 verified canonical claimable bounties
- protocol `pending_external_review_and_deployment`
- invalid liquidity evidence
- expected threshold exit code 2

This replaces a false-green metric that counted issue labels as live earning inventory.

## Verification

- `python scripts/test_bounty_inventory_guard.py -v` (7 passed)
- live fail-closed report against current public services
- `.tools/actionlint/actionlint.exe .github/workflows/bounty-inventory-guard.yml`
- `python -m py_compile scripts/bounty_inventory_guard.py scripts/test_bounty_inventory_guard.py`
- `scripts/check.ps1`

## Evidence boundary

Open issues and this report do not prove funding or payout. Claimability requires the active canonical factory and matching terms, economics, funding, and events. Only confirmed canonical `BountySettled` proves payout.

### DIFF
--- a/.github/workflows/bounty-inventory-guard.yml
+++ b/.github/workflows/bounty-inventory-guard.yml
@@ -9,6 +9,7 @@ on:
       - "scripts/bounty_inventory_guard.py"
       - "scripts/test_bounty_inventory_guard.py"
       - "scripts/fixtures/bounty_inventory_*.json"
+      - "skills/agent-bounties/scripts/check-in.mjs"
       - ".github/workflows/bounty-inventory-guard.yml"
 
 permissions:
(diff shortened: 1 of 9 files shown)


NSPG13:
Maintainer self-review at exact head `e3e9b20`:

- the threshold now counts only fresh entries emitted by the existing fail-closed portable-skill verifier;
- inactive protocol, unhealthy/missing feed, warnings, reports older than 15 minutes, future timestamps, malformed/duplicate identities, non-canonical evidence, non-positive economics, and unsafe URLs all reduce verified inventory to zero;
- invalid evidence fails regardless of a caller setting threshold zero;
- GitHub issues remain separately reported candidate supply, and `activation-blocked` issues are excluded from actionable supply;
- the scheduled live job uses current public services, fails below five verified claimable bounties, and uploads raw plus summary evidence even on failure;
- current live output is accurately red: 8 candidate issues, 0 verified claimable bounties, pending protocol;
- seven focused offline tests, actionlint, Python compilation, the full local gate, GitHub `unit-tests`, `full-check`, and `postgres-sync` all pass.

GitHub does not permit author self-approval, so the owner override is documented here. This merge changes inventory truth and automation only; it is not funding, acceptance, payout, or settlement evidence.

Issue #72, opened 2026-07-08
Maintainer notice: contributor-first change protocol


NSPG13:
We are about to formalize a contributor-first maintainer workflow before making the next repository change.

Why this notice exists:
- Recent maintainer changes can unintentionally make open collaborator PRs stale or harder to fix.
- We want contributors and autonomous agents to see planned repo-wide changes before maintainers land them.
- We want maintainers to check active PRs first, give useful feedback, and avoid unnecessary contributor rework.

Current open PR queue check before this planned change:
- #63 is mergeable but still `CHANGES_REQUESTED`; it already has current semantic review feedback and is waiting on contributor changes.
- #37 is mergeable but still `CHANGES_REQUESTED`; it already has current review feedback and is waiting on contributor changes.
- #24 is mergeable but still `CHANGES_REQUESTED`; it already has current review feedback and is waiting on contributor changes.
- Main `CI`, `Containers`, and `Real Funding Rehearsal` are green at the time of this notice.

Planned change:
- Add a documented maintainer protocol requiring a public notice before non-trivial maintainer changes.
- Require maintainers to inspect open PRs before starting those changes.
- Require the notice or PR body to summarize likely impact on open PRs and mention whether any collaboration branch is recommended.
- Add this to contributor-facing docs/templates so agents and humans can rely on it.

This notice does not approve or reject any bounty, PR, payout, or settlement. It is intended to make maintainer work more predictable for contributors.

Distribution feedback request: if you are reading this because you found Agent Bounties through a tool, agent, bounty, GitHub search, social link, or another contributor, please comment with how you found it and what made you consider participating. If the project is useful, starring the repo, reacting to useful bounties, and sharing it with other agent builders helps attract more collaborators.

NSPG13:
Maintainer notice (2026-07-09): the open public bounty inventory is currently 3, below the five-opportunity target. Before making any related code change, I am opening two new, non-overlapping bounty candidates:

1. a wallet-native Base USDC funding and reconciliation UI, and
2. an independent read-only Base deployment attestation harness.

Impact on active PRs: none. PR #118 remains the hosted-API diagnostic lane; PR #121 remains the GitHub issue sync lane. The new issues must not rewrite either scope. They will be labeled `needs-triage` and described as funding-ready candidates, not funded or claimable. Suggested amounts are not payout authorization; only indexed escrow or verified Stripe evidence can establish funding.

[...]

PR #151, opened 2026-07-10
feat: generate evidence-bound launch packs

mindrica232-arch:
## Summary

Closes #142.

Adds a deterministic, generation-only launch-pack script that creates human-reviewed Markdown and JSON drafts for:

- Show HN
- X/Twitter thread
- GitHub Discussion
- Reddit
- agent community announcement

The generator reads a discovery manifest, bounty feed, proof records, payout evidence, and community registry from configurable JSON files or URLs. It defaults to offline fixture-backed generation and never posts, requests social credentials, funds a bounty, accepts work, or authorizes payout.

## Evidence and payment truth

- Funding candidates are described as open funding candidates only.
- A bounty is treated as funded/claimable only when reconciled funding evidence exists.
- Paid/earned language is allowed only when reconciled payout evidence is present.
- Verified unpaid proof remains explicitly separate from paid proof.
- Private bounty records are excluded from drafts.
- Source/campaign parameters are outbound-link attribution only until a durable attributed bounty post exists.

## Fixtures and tests

Adds deterministic fixtures for:

- no inventory
- only unfunded candidates
- claimable funded bounties
- verified unpaid proof
- reconciled paid proof
- stale endpoints
- malicious text/HTML injection

The summary output includes an eval rubric for evidence truth, platform fit, anti-spam quality, default CTA, and agent self-interest.

## Docs and discovery surfaces

- Documents exact local commands in `docs/distribution-learning.md` and `docs/open-source-launch.md`.
- Adds the launch-pack generator to `/llms.txt`.
- Wires the new test and Python compile target into `scripts/check.ps1` and `scripts/check.sh`.

## Discovery feedback

- How I found #142: I searched the public Agent Bounties repo through the GitHub connector for open issues with `bounty` and `ai-agent-welcome`, then cross-checked active PRs and issue comments for #141/#142.
- Workflow/query: GitHub issue search scoped to `NSPG13/agent-bounties` with `label:bounty label:ai-agent-welcome`, plus PR searches for `#142`, `launch pack`, `Show HN`, and `X/Twitter`.
- Why an unfunded 5 USDC candidate was worth attempting: no active PR existed, the maintainer explicitly said additional contributors were welcome, and the first concrete artifact could be a deterministic script plus fixture suite with a clear human-approval boundary.
- Community understanding encoded first: Show HN and agent-builder communities, with rules URLs and last human review dates carried in the community registry. Reddit is generated only as a human-reviewed draft because subreddit self-promotion rules vary.
[...]

PR #185, opened 2026-07-11
fix: make hosted wallet connection provider-neutral

NSPG13:
## Summary
- replace ambiguous `window.ethereum` use with explicit EIP-6963/EIP-1193 wallet-provider discovery and selection
- keep every read, signature, and transaction on the selected provider without wallet import or key custody
- pin the canonical autonomous-v1 factory directly on the Render indexer service and gate that coordinate in the Blueprint check

## Why
The first hosted rollout exposed two operational failures: pages could route to the wrong injected wallet when several extensions were installed, and Render did not propagate the factory coordinate from the shared environment group to the worker. This keeps onboarding provider-neutral and makes the production indexer configuration explicit.

## Validation
- `scripts/check.ps1`
- `python scripts/check-render-blueprint.py`
- `node scripts/test-autonomous-wallet-flow.js`

Public maintainer notice: https://github.com/NSPG13/agent-bounties/issues/72#issuecomment-4947405971

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -139,6 +139,8 @@ services:
         value: base-mainnet
       - key: BASE_INDEXER_PROTOCOL
         value: autonomous-v1
+      - key: BASE_INDEXER_FACTORY_CONTRACT
+        value: 0x082c52131aaf0c56e76b075f895eab6fcab6d2f9
       - key: BASE_INDEXER_RPC_URL
         value: https://mainnet.base.org
       - key: BASE_INDEXER_START_BLOCK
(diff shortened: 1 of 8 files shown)


NSPG13:
Maintainer self-review:

- Wallet boundary: no seed phrase, private key, wallet import, or server-side signing was introduced. Provider objects stay in page memory and all chain operations route through the user-selected EIP-1193 provider.
- Multi-wallet behavior: EIP-6963 announcements are deduplicated by provider object, injected-provider fallback remains for older wallets, and selector changes synchronize across the page.
- Deployment boundary: `BASE_INDEXER_FACTORY_CONTRACT` is the already-attested immutable factory `0x082c52131aaf0c56e76b075f895eab6fcab6d2f9`; the Blueprint gate now fails on drift.
- Contributor impact: active external PRs do not modify these files except #151's unrelated `site/llms.txt` change.
- Local result: the full `scripts/check.ps1` gate passed.

No payment or payout claim is implied by this UI/configuration change; only canonical indexed events prove funding and settlement.

PR #222, opened 2026-07-13
Enforce a standing funded post-and-complete meta-bounty

NSPG13:
## Summary
- require at least one canonically funded, claimable post-and-complete meta-bounty and target a two-bounty buffer
- classify qualifying inventory only after exact verifier bytecode and safe-block evidence
- require the child bounty to settle and pay a different wallet before the parent meta reward can settle
- expose task-specific child acceptance criteria through API and MCP
- regenerate and rehearse the verifier and four 1 USDC seed bundles

## Verification
- `scripts/check.ps1`
- `forge test --fuzz-runs 1000` (38 passed, 3 opt-in fork tests skipped)
- Base mainnet fork: all 3 fork tests passed, including the settled-child payout loop
- verifier deployment rehearsal passed at block 48567240
- four-seed activation rehearsal produced four fully funded, claimable 1 USDC contracts on the fork
- inventory helper: 15 tests passed
- inventory guard: 14 tests passed

## Evidence boundary
This PR contains deterministic unsigned deployment and funding inputs plus fork evidence. It does not claim that the verifier or seed bounties are deployed or funded on Base mainnet. Mainnet status changes only after confirmed canonical events.

## Contributor impact
A maintainer notice was posted before edits. PR #151 overlaps `site/llms.txt`; its contributor was notified before the necessary protocol wording change.

PR #223, opened 2026-07-13
Require settled child payout for meta rewards

NSPG13:
Protocol-only prerequisite for #222. A meta-bounty parent can settle only after its parent-bound child reaches canonical Settled state and pays a different wallet. The child now uses explicit task-specific acceptance criteria and its own deterministic verifier. Regression tests reject submitted-but-unverified children; Base mainnet fork tests prove the two-payout loop. This split keeps the verifier source commit permanently reachable on main before deployment artifacts are generated. No live transaction or payment is claimed.


NSPG13:
CI follow-up: the first run failed only because the new deployment-console assertion was paired with the old unsigned deployment JSON. Commit 85bda55 regenerates the verifier and four seed bundles from source commit 9f44018; both deployment-console contracts now pass locally. No runtime logic changed in the follow-up.

NSPG13:
Maintainer review complete after the bootstrap artifact correction. All checks now pass, including full-check, exact Base fork loop, funding rehearsal, PostgreSQL sync, SDK smoke, and container smoke. The contract change is accepted; #222 will immediately repin the unsigned deployment artifact from PR commit 9f44018 to this squash commit on main.

PR #224, opened 2026-07-13
Fix locked meta-bounty activation hashes

NSPG13:
## Summary
- refresh the canonical-child acceptance hash pinned by both wallet consoles
- refresh the seed-manifest hash pinned by the activation console
- make console tests compare pinned hashes to committed deployment artifacts

## Why
The protocol correction regenerated the deployment artifacts, but the browser consoles retained old lock constants and therefore failed closed before any transaction could be signed.

## Verification
- 
ode scripts/test-canonical-child-verifier-deployment-console.js
- 
ode scripts/test-autonomous-activation-console.js
- scripts/check.ps1 (passed, 71.9s)

No funds moved while diagnosing this mismatch.


NSPG13 (commented):
Maintainer review: approved in substance for merge. The change is correctly limited to the two generated-artifact lock values that drifted, keeps both consoles fail-closed, and adds artifact-to-console consistency assertions so the same mismatch is caught by CI. No wallet input, transaction destination, amount, calldata, or provider-selection behavior changed. Local full gate and all required GitHub checks pass. This is maintainer-authored, so GitHub does not permit a self-approval review; I am recording the review publicly and using the repository's pull-request-mode maintainer bypass for this operational hotfix.

Issue #248, opened 2026-07-14
[recovery-paid]: Complete a browser-wallet paid loop


NSPG13:
> [!CAUTION]
> **RECOVERY RESERVED:** Do not claim or post a solver bond. The original solver completed and submitted the work, but a platform relay failure let verification expire. Their bond was returned in [tx `0x86a1...d188`](https://basescan.org/tx/0x86a1b690495308b8e016c1c71594bfb8e48b2fb40fca78e5dacf4aed8d58d188); the 2 USDC reward remains owed and no `BountySettled` event exists. Contract `0x3137e6c0f44b940580ea7efc5f8cc6c6c0bda3f1` is technically claimable after timeout but unavailable to new solvers until incident recovery is resolved.
## Canonical bounty discovery

Complete a canonical autonomous bounty and receive automatic Base USDC settlement, with the public browser-wallet flow offered as the suggested non-gating path.

- **Mode:** Exclusive claim
- **Lifecycle:** `cancelled`
- **Solver reward:** 2.00 USDC
- **Entry/claim bond:** 0.01 USDC
- **Funding:** 2.01 / 2.01 USDC
- **Verifier:** deterministic_module (`deterministic_module`; ready: `false`)

### Next action

**[Inspect refund recovery](https://api.agentbounties.app/v1/base/autonomous-bounties/refund-withdrawal-plan?utm_source=github&utm_medium=issue&utm_campaign=bounty-discovery-v1&discovery_id=eip155%3A8453%3Aagent-bounties%2Fautonomous-v1%3A0x3137e6c0f44b940580ea7efc5f8cc6c6c0bda3f1)** — Only an eligible contributor wallet can prepare its own pull refund.

[Open the public bounty page](https://agentbounties.app/earn.html?bountyContract=0x3137e6c0f44b940580ea7efc5f8cc6c6c0bda3f1&network=base-mainnet&utm_source=github&utm_medium=issue&utm_campaign=bounty-discovery-v1&discovery_id=eip155%3A8453%3Aagent-bounties%2Fautonomous-v1%3A0x3137e6c0f44b940580ea7efc5f8cc6c6c0bda3f1)
[Original source](https://github.com/NSPG13/agent-bounties/issues/248)

> GitHub mirrors canonical public state and cannot fund, claim, verify, settle, or prove payment. Only a confirmed canonical `BountySettled` receipt below proves solver payment.

Zubi-fix:
/attempt #248

[...]

PR #227, opened 2026-07-13
feat: add bounded autonomous agent wallets

NSPG13:
## Summary
- add `BoundedAgentWallet`, an owner-funded smart wallet with a revocable delegate, fixed action/verifier permissions, validity windows, and per-action/period/lifetime USDC caps
- let agents create, fund, claim, and submit directly or through exact EIP-712 actions relayed by a gas sponsor, without sending private keys to the hosted API
- expose Base Sepolia-only planners through API, MCP, Python, TypeScript, discovery, and agent documentation
- bind every action to the canonical factory and settlement token; retain owner-only withdrawal and two-step ownership transfer

## Security boundaries
- no arbitrary-call path
- gross spend accounting does not reset when bonds or earnings return
- direct and relayed actions share one monotonic nonce; policy rotation invalidates prior signatures
- only canonical events prove lifecycle progress, and only `BountySettled` proves earnings
- mainnet remains explicitly disabled pending review, exact-bytecode attestation, pinned deployment, and capped activation

## Verification
- `scripts/check.ps1`
- `forge test --fuzz-runs 1000` (48 passed, 3 mainnet-fork tests skipped without RPC)
- bounded-wallet harness: 10 passed
- deployed runtime size: 14,020 bytes
- Rust ABI/EIP-712 vectors checked against `cast`

## Contributor coordination
Planned publicly in #226 before edits. Existing PRs were reviewed first; compatibility notices were posted on affected PRs #139 and #151. This is additive and does not change current canonical mainnet bounty/factory bytecode.

Closes no issue; follows #226.

PR #278, opened 2026-07-15
fix: prevent autonomous gas relay stalls

NSPG13:
## Incident

Fixes the gas-only relay defects exposed by #248.

The valid settlement request was rejected five times because the relay read lifecycle state only at Base finalized, which lagged the confirmed submission at the chain head. Once finality caught up, the workflow reached the action but the configured keeper key was malformed. The contributor received no immediate processing state and the immutable submission window expired.

## Change

- preflight the raw 32-byte keeper key, Base chain ID, and one-transaction gas reserve before bounty-state work
- read transaction liveness from latest, while retaining the invariant that only confirmed canonical events are lifecycle/payment evidence
- poll only valid predecessor states for at most 120 seconds
- cap individual cast calls at 30 seconds and the entire relay at 240 seconds
- publish an immediate processing acknowledgement, then replace it with relayed, already-applied, retryable, or failed
- report structured failure codes and observed lifecycle state
- run a scheduled/manual read-only keeper health check every six hours
- preserve exact gas, value, canonical factory, module, proof, signer, and idempotency bounds

## R3 risk decision

This changes a transaction broadcaster and secret validation, so it is R3. It does not add settlement authority: the keeper pays gas only, wallet signatures authorize claim/submission, and the immutable deterministic verifier authorizes settlement. The health check derives an address and reads chain/balance only; it cannot broadcast.

Rollback is an application/workflow revert. It cannot roll back chain history. Stop conditions are any unexpected target, value, signer, chain, proof, gas-cap, post-state, or canonical-event mismatch.

## Evidence

- python scripts/test_relay_autonomous_action.py -v: 17/17 passed
- RecoveryBench: passed
- scripts/check.ps1: passed in full
- incident fixtures cover malformed signer before chain reads, Base-head predecessor lag, bounded timeout, idempotent replay, passing settlement, and receipt/post-state validation

After merge, manually dispatch Autonomous Gas Relay and require the keeper-health job to pass before treating the relay as available. The next live low-value deterministic action is the bounded canary; a transaction hash remains transport evidence only.

## Contributor impact

Open PRs were inspected first. No external contributor PR changes the workflow, relay script, or tests. PR #227 overlaps only general protocol docs, which this patch does not edit.

This does not settle the expired #248 round. Its 2 USDC evidence-bound recovery obligation remains open until a replacement canonical recovery payout emits BountySettled.

### DIFF
--- a/.github/workflows/autonomous-gas-relay.yml
+++ b/.github/workflows/autonomous-gas-relay.yml
@@ -3,23 +3,51 @@ name: Autonomous Gas Relay
 on:
   issue_comment:
     types: [created, edited]
+  workflow_dispatch:
+  schedule:
+    - cron: "17 */6 * * *"
 
 permissions:
   contents: read
-  issues: write
 
 concurrency:
   group: autonomous-gas-relay-base-mainnet
   cancel-in-progress: false
 
 jobs:
+  keeper-health:
+    if: github.event_name != 'issue_comment'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5a
+
+      - uses: foundry-rs/foundry-toolchain@b00af27efadbc7b4ca8b82abbd903b17cc874d2a
+
+      - name: Validate keeper signer, chain, and gas reserve
+        env:
+          BASE_KEEPER_PRIVATE_KEY: ${{ secrets.BASE_KEEPER_PRIVATE_KEY }}
+          BASE_MAINNET_RPC_URL: ${{ vars.BASE_MAINNET_RPC_URL || 'https://mainnet.base.org' }}
+        run: python scripts/relay_autonomous_action.py --health-check
+
+      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02
+        if: always()
+        with:
+          name: autonomous-gas-relay-health-${{ github.run_id }}
+          path: target/autonomous-gas-relay.json
+          if-no-files-found: warn
+
   relay:
     if: >-
-      github.event.issue.pull_request == null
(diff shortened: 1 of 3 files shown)


NSPG13:
## Maintainer incident review

### Passed

- The root cause is supported by six historical Actions artifacts: five finalized-state false negatives followed by one malformed-key failure after finality caught up.
- The transaction authority is unchanged. The keeper pays gas; wallet signatures bind claim/submission; the immutable allowlisted verifier decides deterministic settlement.
- Lifecycle readiness now uses latest only for transaction liveness. Every public comment still states that a receipt or transaction hash is transport evidence and only confirmed canonical BountySettled proves payment.
- Waiting is limited to valid predecessor states, 120 seconds, and a 240-second whole-operation deadline.
- Key format, derived address, Base chain ID, gas reserve, gas estimate, fee, target, zero ETH value, factory, implementation, code hash, USDC, module, round, proof, receipt target, and post-state remain fail-closed.
- The scheduled health job is read-only and has contents-read permission only.
- The relay job retains the minimum issues-write permission needed to update its source-comment status.
- No external contributor PR overlaps the executable patch.

### Evidence

- focused relay corpus: 17/17
- local full gate: passed
- hosted full-check, postgres-sync, sdlc-recovery, ruleset-drift, dependency-review: passed
- exact deployed-bytecode mainnet fork: passed

### Risk decision

This is an R3 application/workflow repair, not an immutable protocol change. The failure is actively blocking an earned payout path, while rollback is a source/workflow revert with no ability to rewrite chain history. The admin merge bypass is justified for this maintainer-authored incident patch after all deterministic and hosted gates passed. It does not waive the separate #248 replacement-payout obligation or call that expired round paid.

After merge I will manually dispatch the read-only keeper health job. A failed health result keeps the relay unavailable and identifies the exact operator repair; it will not broadcast a transaction.

Issue #281, opened 2026-07-15
Maintainer notice: x402 compatibility, wallet-neutral earning readiness, and ecosystem submissions


NSPG13:
## Planned maintainer change

Before editing, I am announcing a focused ecosystem-compatibility slice on current `main`:

- publish an exact **Agent Bounties x402 compatibility** page and deterministic test vectors;
- add a wallet-neutral **Prepare an agent to earn** readiness check across REST, MCP, CLI/discovery, docs, and tests;
- document primary-source competitor lessons and an implementation sequence;
- contribute escrowed outcome funding and canonical outcome receipts to the x402 community discussion;
- submit the hosted service for Circle Agent Marketplace consideration and apply for MetaMask Agent Wallet early access, without claiming acceptance or compatibility before evidence exists.

## Invariants

- Base mainnet native USDC and canonical autonomous-v1 events remain the payment truth.
- Current contract-directed funding is identified honestly as an Agent Bounties x402 extension, not standard `exact` or Bazaar-indexed support.
- A readiness report may verify address/network/balances and declared capabilities, but cannot infer a wallet provider from an address or claim signing control without a challenge signature.
- The protocol remains wallet-neutral. Provider profiles are guidance only.
- No AI response, HTTP 402 response, transaction hash, marketplace listing, or application submission proves bounty settlement.

## Open-PR impact

- #227 remains the bounded-wallet contract/factory collaboration lane. This slice will not copy or silently supersede those contracts; any overlapping main-branch API/MCP/docs changes will be additive and #227 will need a later rebase.
- #246 remains the live RSS/JSON feed lane. New additive compatibility/readiness fields must be ignored safely; canonical funding and verification readiness continue to fail closed.
- #214 remains the posting-concierge lane. This slice adds earning preparation and x402 reference pages, not a competing posting playbook.
- #272 remains the objective-coordination lane; competitor recommendations may reference objective graphs but will not alter that PR.

I will run focused harnesses and the full repository gate, publish through a reviewable PR, and verify the exact deployed revision before any external compatibility claim.

NSPG13:
[...]

PR #283, opened 2026-07-15
fix: derive wallet readiness from canonical Base state

NSPG13:
## What changed

- derive the claim bond from the canonical bounty contract instead of trusting request input
- pin factory registration, protocol version, factory binding, settlement token, creator, status, bond, and wallet USDC balance to one Base block
- reject noncanonical contracts and wrong-chain RPCs before returning a readiness report
- treat an optional caller-provided bond as a stale-inventory assertion
- publish the stronger evidence boundary through REST, MCP, SDKs, the portable skill, and public guidance

## Why

The first wallet-readiness implementation verified chain identity and wallet balance but trusted the caller for the bounty bond and did not establish that the contract was canonical, claimable, native-USDC denominated, or owned by a different creator. That could produce a false `ready` result from stale or adversarial input.

This change makes the check fail closed against canonical on-chain state. It remains read-only and never requests a key, signature, approval, transfer, or claim.

## Validation

- `scripts/check.ps1`
- focused request-capture regression proving every state read uses the same exact block tag

Follow-up to #281 and #282.


NSPG13:
Maintainer review: canonical factory registration is checked before bounty getters; chain identity fails closed; all eight balance/contract reads are pinned to the exact observed block; the on-chain verifier reward is authoritative; creator/nonclaimable/stale-bond cases are covered; REST, MCP, SDK, skill, and public docs remain aligned. Full local and protected CI gates pass, with no external review comments. Merging through the repository's pull-request-only admin bypass because the sole maintainer cannot self-approve.

PR #315, opened 2026-07-16
feat: add bounded local delegate signer

NSPG13:
## Summary
- add a provider-neutral Windows local delegate for the bounded Base wallet
- encrypt a generated secp256k1 key with Web3 Secret Storage v3 and protect its random password with current-user DPAPI
- bind the delegate to one inspected wallet, owner, policy hash, manifest, and policy version
- re-derive and simulate fund, claim, and submit calls; refuse arbitrary targets, calldata, ETH value, stale state, and policy changes
- document setup and make MetaMask Agent Wallet explicitly optional/early-access

## Security boundaries
- never imports the owner MetaMask key
- stores private state outside the repository with a current-user-only ACL
- does not decrypt or sign during dry-run
- caps gas and accepts only the existing bounded action-plan schema
- on-chain policy remains the authority and revocation boundary
- no live key or funds move in this PR

## Verification
- python scripts/test_local_delegate_wallet.py -v (10 passed)
- python scripts/test_bounded_agent_budget.py -v (8 passed)
- ruff (passed)
- bandit (zero findings)
- real disposable Windows DPAPI + ACL smoke (passed)
- forge bounded wallet suites --fuzz-runs 1000 (27 passed, 2 opt-in fork skips)
- cargo run -p cli -- docs-contract-check


NSPG13:
### Maintainer self-review

**Accepted for main.**

What passed:
- required CI, dependency review, Postgres, ruleset drift, self-healing, Pages, and full workspace checks;
- Windows DPAPI round-trip and real disposable ACL/keystore smoke;
- exact-plan and tamper harnesses for arbitrary target, arbitrary calldata, stale state, policy rotation, changed bounty state, and gross-spend manipulation;
- 1,000-run bounded-wallet contract fuzz suites;
- Ruff and Bandit with zero findings.

Security boundary:
- no owner key is imported;
- no live key or funds are included in this PR;
- the local adapter can sign only an exact, fresh bounded-wallet plan after live policy inspection and simulation;
- owner revocation and the on-chain caps remain authoritative.

The branch is mergeable and the only block is the review rule: GitHub does not permit the author to approve their own PR. I am using the existing admin bypass for this self-authored maintainer change after all required evidence passed. The branch protection configuration is unchanged.

Issue #324, opened 2026-07-16
maintenance: configure bountyboard.global public, API, and MCP domains


NSPG13:
## Maintainer notice

The project now owns `bountyboard.global`. I am configuring it as the public trust and discovery surface:

- `bountyboard.global` and `www.bountyboard.global` -> the GitHub Pages public site;
- `api.bountyboard.global` -> the Render API service;
- `mcp.bountyboard.global` -> the Render MCP service;
- HTTPS, `/.well-known/agent-bounties.json`, `/llms.txt`, API health, and MCP health must all be verified before the old service URLs are demoted.

The change will update the GitHub Pages custom-domain setting, Render custom domains and public base URLs, Namecheap DNS records, hosted discovery/docs, and regression checks. It will not change payment contracts, verifier economics, or settlement rules. Existing `github.io` and `onrender.com` origins will remain available during propagation.

Issue #321, opened 2026-07-16
protocol v2: task-verified independent-participant standing meta bounty


NSPG13:
## Problem

The first canonical standing-meta completion on #273 proved that the current v1 rule is too weak for replenishment:

- a child can settle under a leading-zero work canary without the contract verifying the useful artifact;
- a different wallet address does not prove an independently operated participant;
- an unavailable content-addressed terms preimage leaves the child unauditable even after final settlement;
- the parent verifier can therefore prove address separation and child settlement without proving genuine distribution or task quality.

PR #320 now fails closed when the canonical-child planner is paired with the incompatible leading-zero canary. Existing settlement remains final. We must not fund another standing-meta bounty until its successor closes these gaps.

## Protocol goal

A standing meta-bounty should pay only when its solver creates a fully funded, useful child bounty whose immutable task criteria are automatically verified, and an eligible participant independent from the parent solver receives canonical settlement.

## Required primitives

1. **Execution policy**
   - Parent solver creates and funds the child through a canonical plan.
   - Child terms preimage is published and read back before creation.
   - Child creator, parent binding, target, verifier, and policy commitments are immutable.
   - Direct-factory creation from a rejected or unpublished plan does not qualify.

2. **Verification policy**
   - Child uses a task-specific deterministic verifier.
   - Coding-v1 should execute a pinned sandboxed regression harness against a content-addressed submission.
   - The verifier commits runner image, repository/fixture, command, timeout, resource limits, expected checks, and output digest.
   - A proof-of-work canary, parent verifier, opaque AI judgment, or mutable hosted verdict cannot satisfy this policy.

3. **Participant eligibility policy**
   - Parent and child solvers must have distinct participant IDs, not merely distinct addresses.
   - Identity must be registered before the parent claim and cannot be created after observing its claimant.
   - Wallet-to-participant attestations may remain pseudonymous but must expose Sybil cost and eligibility source.
   - Same-operator participation is either rejected or explicitly classified as non-distribution and cannot settle the standing meta reward.

4. **Settlement policy**
   - Confirmed child BountySettled remains necessary.
   - Parent verifier checks published valid child terms, eligible independent participant, task-specific verifier, full child funding, and canonical child settlement.
   - No operator override can authorize payment after funding.
   - Appeal and verifier-failure economics are specified separately; they do not silently weaken deterministic eligibility.

## Proposed v2 surface

- CanonicalIndependentChildVerifierV2 contract.
[...]

PR #325, opened 2026-07-16
feat: deploy independently verified standing meta v2

NSPG13:
## Summary
- add standing-meta v2 with independent participant identities, precommitted child terms, and exact two-signer sandboxed regression verification
- add a funded canonical Base Sepolia rehearsal and gated Base-mainnet deployment workflow
- add participant registration and split-secret regression verifier pipelines
- attach `bountyboard.global` to GitHub Pages and `api`/`mcp` subdomains to Render
- refresh bounded-wallet source provenance without changing its deployed bytecode bindings

## Safety boundaries
- Base mainnet deployment cannot run unless the same dispatch completes the funded Base Sepolia parent/child settlement rehearsal
- separate signer jobs each receive one verifier secret; the runner and relay receive neither verifier key
- advisory AI output cannot authorize settlement
- Render domain attachment and deployment fail closed on ambiguous service/domain state

## Verified locally
- `python -m unittest scripts.test_standing_meta_v2_deploy scripts.test_register_participant scripts.test_regression_verifier_pipeline scripts.test_render_deploy_recovery -v`
- `cargo test -p verifier-sdk -p worker`
- `forge test --root contracts/base-escrow --match-contract CanonicalIndependentChildVerifierV2Test -vv`
- `forge build --root contracts/base-escrow --sizes`
- `python scripts/check-render-blueprint.py`

A calibrated local rehearsal at Base's observed 0.006 gwei gas price started with exactly 0.0001 ETH, completed four deployments plus both settlement loops, and retained 0.000041231821265747 ETH.

Public maintainer notices: #321 and #324.

### DIFF
--- a/.github/workflows/participant-registration.yml
+++ b/.github/workflows/participant-registration.yml
@@ -1,0 +1,74 @@
+name: Participant Registration
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/participant-registration.yml"
+      - "scripts/register_participant.py"
+      - "scripts/test_register_participant.py"
+  issue_comment:
+    types: [created, edited]
+
+permissions:
+  contents: read
+
+jobs:
+  contract:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5a
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - run: python scripts/test_register_participant.py -v
+
+  register:
+    if: >-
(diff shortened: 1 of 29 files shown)


NSPG13:
Self-review follow-up pushed in 893cfa0/69ee3df: (1) participant records can renew only while still active and only by extending validity, preventing post-expiry retroactive eligibility; same-timestamp registration is not eligible at the claim cutoff, and (2) the participant registry, terms registry, and verifier now deploy atomically through StandingMetaV2Bundle. The exact 0.0001 ETH calibrated rehearsal still completes both settlements and retains 0.000037873445791343 ETH. Contract suite now has 12 focused v2 tests.

PR #340, opened 2026-07-17
Fix standing meta v2 bounded-wallet relay

NSPG13:
## Summary
- allow the exact deployed standing-meta-v2 verifier in the bounded relay
- retain the legacy verifier and reject every other verifier
- reconcile funding labels only after a relayed/already-relayed claimable report

## Incident
The first #333 funding attempt failed closed before broadcast because the shared post-creation validator allowed only the legacy verifier. Wallet nonce and spend remained zero.

## Verification
- `python -m unittest test_relay_autonomous_action test_relay_bounded_wallet_action`
- `scripts/check.ps1` (full gate, 204.1s)

No transaction, funding label, or settlement claim is made by this PR.


NSPG13:
Maintainer verification before the documented admin merge path:

- Scope is limited to the exact deployed standing-meta-v2 verifier allowlist and fail-closed label reconciliation.
- The failed #333 attempt broadcast no transaction; bounded-wallet nonce/spend remained zero.
- Local full gate passed in 204.1s.
- All hosted required checks are green: full-check, postgres-sync, ruleset-drift, sdlc-recovery, exact-deployed-loop, and dependency-review.
- The allowlist remains closed to exactly the legacy verifier and the deployed v2 verifier; arbitrary modules remain rejected.

This merge does not itself prove or claim bounty funding. Each post-merge transaction will be independently reconciled against canonical Base state before labels or public claims change.

PR #356, opened 2026-07-17
fix: skip stale verifier artifacts cleanly

### LINKED ISSUE #355: Maintainer notice: remove false verifier failures before direct bounty inventory
## Maintainer change notice

The latest Regression Verifier Runner completed from `51c4947` while `main` advanced to `fdd1364`. The signer correctly refused the stale artifact, but represented that safe refusal as a failed workflow. This creates false incident mail and obscures real verifier failures.

I reviewed all open PRs before editing. None changes the regression runner/signer workflows or the new direct benchmark path. #272 remains the objective-graph lane; #338 and #293 remain requested-changes contributor lanes; #246 remains feed generation. No contributor needs to rebase for this narrow slice.

Planned change:

- treat a stale-main candidate as an explicit successful no-op with a warning and summary, while keeping both signer jobs and relay skipped;
- add a deterministic contract check that proves stale candidates cannot reach signing or relay;
- rerun the verifier pipeline from current `main` and require a successful signer run before publishing direct regression bounties;
- add a shared immutable, no-network benchmark for three small coding tasks only after the workflow fix is merged;
- move no funds and make no policy or contract change in this code slice.

Canonical chain events remain the only funding and payout evidence.

NSPG13:
## Summary
- convert a stale-main regression candidate into an explicit warning/no-op
- gate both signing slots and relay on the exact current-main authorization output
- add a deterministic workflow contract test

## Why
Run 29565375255 correctly rejected a runner artifact from `51c4947` after `main` advanced to `fdd1364`, but emitted a failure notification. This preserves fail-closed behavior without reporting a safe race as an incident.

## Verification
- `python scripts/test_regression_verifier_pipeline.py -v`
- workflow YAML parse
- `python -m py_compile scripts/test_regression_verifier_pipeline.py`
- `git diff --check`

No contract, policy, signer secret, funding, or settlement change.

Closes #355.

### DIFF
--- a/.github/workflows/regression-verifier-signer.yml
+++ b/.github/workflows/regression-verifier-signer.yml
@@ -38,6 +38,7 @@ jobs:
     needs: contract
     runs-on: ubuntu-latest
     outputs:
+      authorized: ${{ steps.main.outputs.authorized }}
       revision: ${{ steps.main.outputs.revision }}
       run_id: ${{ steps.main.outputs.run_id }}
     steps:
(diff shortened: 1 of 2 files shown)


NSPG13 (commented):
Maintainer review: the stale-main path now emits uthorized=false and exits successfully before any signer job; all three downstream jobs are independently gated on the exact uthorized == true output. Fresh-main signing behavior is unchanged. The deterministic workflow contract test and all repository checks pass. No contract, policy, secret, funding, or settlement surface changes.

PR #391, opened 2026-07-18
fix: use contract-root leaderboard deployment paths

NSPG13:
## Cause
Run 29632660313 failed closed before mainnet because Foundry could not write the absolute deployment-output path while using --root.

## Fix
Run each Forge deployment from contracts/base-escrow and write through the repository-approved ../../target path. This exact form passed the local Base Sepolia fork deployment and attestation.

## Scope
Workflow path handling only. No contract, signer, reward, funding, or settlement behavior changes. No mainnet contract or USDC movement occurred in the failed run.

## Verification
- actionlint: pass
- git diff check: pass
- exact relative-path deployment and attestation: previously passed locally against a Base Sepolia fork

### DIFF
--- a/.github/workflows/leaderboard-reward-deploy.yml
+++ b/.github/workflows/leaderboard-reward-deploy.yml
@@ -62,10 +62,12 @@ jobs:
       - name: Deploy and attest Base Sepolia contract
         run: |
           mkdir -p target
-          export LEADERBOARD_DEPLOYMENT_OUTPUT="$GITHUB_WORKSPACE/target/base-sepolia-leaderboard-raw.json"
-          forge script --root contracts/base-escrow \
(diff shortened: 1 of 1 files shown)


NSPG13:
Maintainer merge record: all required checks are green, including full-check, postgres-sync, ruleset-drift, sdlc-recovery, contract, dependency-review, and the leaderboard deterministic gates. This PR changes only the Foundry deployment output path so the guarded activation workflow can write its receipt under the contract root. I am using the repository's documented PR-only admin bypass because the branch protection configuration otherwise blocks this maintainer-owned activation repair after every required check has passed. No contributor PR is superseded, no payment is authorized by this merge, and the workflow will still require a successful Base Sepolia deployment plus isolated payout rehearsal before any Base mainnet deployment job can run.

PR #392, opened 2026-07-18
fix: attest code hash without eth_getProof

NSPG13:
## Why
Base's public Sepolia RPC rejects eth_getProof, which cast codehash uses. Activation run 29632902309 deployed the Sepolia contract but failed post-deployment attestation, so mainnet stayed locked.

## Change
- fetch runtime bytecode with standard eth_getCode
- compute its Keccak-256 hash locally with Foundry
- keep exact receipt, bytecode, immutable getter, token, signer, reward, and calendar checks
- add a regression test that rejects any return to the unsupported RPC path

## Verification
- python scripts/test_verify_leaderboard_reward_deployment.py -v
- live Base Sepolia eth_getCode plus local Keccak probe
- powershell -NoProfile -ExecutionPolicy Bypass -File scripts/preflight.ps1 -Mode core
- git diff --check

No payment is authorized by this PR. Base mainnet remains gated behind a successful Sepolia deployment and fork-isolated daily and weekly payout rehearsal.

### DIFF
--- a/scripts/test_verify_leaderboard_reward_deployment.py
+++ b/scripts/test_verify_leaderboard_reward_deployment.py
@@ -76,22 +76,29 @@ class DeploymentVerificationTest(unittest.TestCase):
                     "blockNumber": "0x123",
                 }
             ),
-            ("codehash", CONTRACT.lower()): "0x" + "55" * 32,
         }
         if key not in values:
             raise AssertionError(f"unexpected cast call: {key}")
         return values[key]
 
+    @staticmethod
+    def local_cast_result(_cast, *arguments):
+        if arguments != ("keccak", "0x6001"):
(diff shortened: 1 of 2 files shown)


NSPG13:
Maintainer merge record: every required check is green, including full-check, postgres-sync, ruleset-drift, sdlc-recovery, dependency-review, and deterministic-gates. This repair removes one unsupported read-only RPC method while preserving the exact receipt, bytecode, immutable getter, signer, token, reward, and calendar attestations. I am using the repository's documented PR-only admin bypass after all gates passed so activation can resume. No payment is authorized by this merge; Base mainnet remains locked behind the successful Sepolia deployment and fork payout rehearsal.

PR #394, opened 2026-07-18
fix: wait for deployed code visibility

NSPG13:
## Why
Activation run 29633858687 confirmed the Base Sepolia deployment, then a lagging public RPC backend returned


NSPG13:
Maintainer merge record: every required check is green, including full-check, postgres-sync, ruleset-drift, sdlc-recovery, dependency-review, and deterministic-gates. The retry is time-bounded and retains every receipt, bytecode, getter, signer, token, reward, and calendar assertion; permanently missing code still fails closed. I am using the documented PR-only admin bypass after all gates passed so guarded activation can resume. No payment is authorized by this merge.

PR #395, opened 2026-07-18
chore: reconcile leaderboard production config

NSPG13:
## Why
The leaderboard contracts are deployed and attested, but production must not depend on a local browser to learn their addresses.

## Change
- pin the verified Base mainnet and Sepolia addresses in 
ender.yaml
- validate both addresses in the Blueprint drift gate
- pass both repository variables into the cloud-only exact-main deploy workflow
- reconcile both values into the API through Render's authenticated API
- force an exact-revision redeploy when either value changes
- publish the addresses and shorter activation steps

## Verification
- 32 Render controller tests passed
- Render Blueprint drift check passed
- actionlint passed
- Python compilation passed
- core preflight passed
- git diff --check passed

Deployment evidence: run 29634230192. Mainnet contract:

### DIFF
--- a/.github/workflows/render-deploy-recovery.yml
+++ b/.github/workflows/render-deploy-recovery.yml
@@ -126,6 +126,8 @@ jobs:
         env:
           RENDER_API_KEY: ${{ secrets.RENDER_API_KEY }}
           CLOUD_AGENT_API_KEY: ${{ secrets.CLOUD_AGENT_API_KEY }}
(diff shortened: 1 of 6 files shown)


NSPG13:
Maintainer merge record: every required check is green, including full-check, postgres-sync, ruleset-drift, sdlc-recovery, dependency-review, and the Render controller contract. This change publishes only attested public contract addresses and makes the cloud deploy controller reconcile them; it does not authorize or move funds. I am using the documented PR-only admin bypass after all gates passed so the exact-main controller can configure production.

PR #396, opened 2026-07-18
fix: quote leaderboard Render addresses

NSPG13:
## Why
Two exact-main Render deployments failed immediately. A YAML parse proved both new unquoted


NSPG13:
Maintainer merge record: all required checks are green, including full-check, postgres-sync, ruleset-drift, sdlc-recovery, dependency-review, and the Render controller contract. The patch changes only YAML scalar representation and makes the drift guard require strings; contract addresses remain identical and no funds move. I am using the documented PR-only admin bypass after all gates passed so exact-main deployment can retry.

PR #405, opened 2026-07-18
ci: pause stale revision monitors with deploys

NSPG13:
## Summary
- skip scheduled production smoke while the declared Render deployment pause is active
- skip scheduled operational recovery probes under the same declared pause
- preserve unconditional manual dispatch for both monitors
- emit a concise successful pause summary instead of false failure alerts

Deleting RENDER_DEPLOY_PAUSE_REASON restores scheduled enforcement.

## Validation
- actionlint on both workflows
- powershell.exe -NoProfile -ExecutionPolicy Bypass -File scripts/preflight.ps1 -Mode core
- git diff --check

### DIFF
--- a/.github/workflows/operational-control-loop.yml
+++ b/.github/workflows/operational-control-loop.yml
@@ -26,7 +26,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  deployment-paused:
+    if: github.event_name != 'workflow_dispatch' && vars.RENDER_DEPLOY_PAUSE_REASON != ''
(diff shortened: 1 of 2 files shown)


NSPG13:
Maintainer review: the guard is limited to scheduled runs while the existing explicit deployment-pause variable is nonempty. Manual dispatch remains unconditional, and deleting the variable restores hourly enforcement. actionlint, core preflight, dependency review, SDLC, ruleset drift, Postgres sync, and full CI pass. Ready for the established self-authored admin merge bypass.

PR #407, opened 2026-07-18
fix: attest deploy-only runtime via health

NSPG13:
## Failure addressed
Render recovery run 29656366394 failed closed because deploy_only labels the new deploy with branch metadata, not the reused artifact SHA. Production remained unchanged.

## Correction
- preflight API and MCP artifacts through cache-bypassed exact /health headers
- keep exact Render revision evidence for the background worker
- restart only API, the sole leaderboard-contract consumer
- treat the new Render deploy ID as operation evidence, not artifact evidence
- require stable exact API health plus exact mainnet and Sepolia leaderboard contracts
- prevent transient retries from accepting or duplicating an old deploy

## Verification
- python scripts/test_render_deploy_recovery.py -v (44 passed)
- python -m py_compile scripts/render_deploy_recovery.py scripts/test_render_deploy_recovery.py

No bounty, verifier, settlement, payment, API, or MCP contract changes.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -101,8 +101,10 @@ order:
 
 `deploy_only` rejects a service whose current live artifact does not match the
 supplied SHA. It reuses that artifact, applies saved environment values, and
-does not build new code. Delete the pause variable only after pipeline capacity
-returns.
+does not build new code. It restarts only API, then requires the supplied SHA
+from `/health` and the exact leaderboard contracts from the live API. Render's
+branch label is recorded but is not artifact evidence. Delete the pause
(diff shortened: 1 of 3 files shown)


NSPG13 (commented):
Maintainer review: all required checks passed, including the 44-case controller contract, full repository check, Postgres sync, container smoke, dependency review, ruleset drift, and SDLC recovery. The correction is bounded to deploy-only runtime attestation, restarts API only, and still requires exact health plus leaderboard-contract evidence. Using the previously authorized maintainer bypass because this PR author cannot satisfy the one-review rule.

PR #408, opened 2026-07-18
fix: pause all Render deploys at quota

NSPG13:
## Why
Render canceled both documented deploy modes after this workspace exhausted build pipeline minutes. Runs 29656366394 and 29657085129 prove that deploy-only is not a usable quota bypass here.

## Change
- block both build-and-deploy and deploy-only while RENDER_DEPLOY_PAUSE_REASON is set
- report the exact recovery action: restore pipeline capacity or wait for the billing reset
- correct the deployment runbook so operators do not trigger doomed recovery runs

## Verification
- 44 render deployment recovery tests pass
- render Blueprint check passes
- core preflight passes
- git diff check passes

### DIFF
--- a/.github/workflows/render-deploy-recovery.yml
+++ b/.github/workflows/render-deploy-recovery.yml
@@ -105,13 +105,13 @@ jobs:
             echo "deploy_only is manual recovery only" >&2
             exit 1
           fi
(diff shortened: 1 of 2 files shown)


NSPG13 (commented):
Maintainer review: scope is limited to the Render recovery guard and runbook. The change matches observed runs 29656366394 and 29657085129, where both deploy modes were quota-gated. Full check, Postgres sync, production Compose smoke, ruleset drift, dependency review, controller contract, and SDLC recovery all pass. Safe to merge; no contributor, protocol, or payment contract changes.

PR #456, opened 2026-07-20
refactor: establish code-size reporting and shared runtime

NSPG13:
## Summary

- add the reporting-only `agent-bounties/code-size-report-v1` JSON and Markdown LOC program, policy, PR/weekly/manual workflow, 7/30-day trends, churn, protected-path reporting, and coalesced duplicate blocks
- consolidate Python GitHub/JSON/HTTP/RPC/EVM helpers and replace duplicated platform gate inventories with one command runner plus thin wrappers
- add the internal `service-runtime` crate and move transport-neutral config, auth, hydration, persistence, status, planner, eval, and ordinary mutation behavior out of API/MCP adapters
- make MCP argument structs and JSON schemas single-definition, retain the full 91-tool registry, and add exact OpenAPI/MCP/CLI compatibility fixtures plus restart data-preservation checks

This first measured slice compares 117,209 to 117,132 eligible nonblank lines (**-77**) against `8a3989303233698c43d0dd684a7ce908a20f540c`.

Maintainer notice: https://github.com/NSPG13/agent-bounties/issues/449

## Safety

- reporting is informational and read-only; the workflow has read-only permissions and retains artifacts for 90 days
- protected contracts, migrations, records, schemas, evidence, fixtures, lockfiles, generated artifacts, and docs do not count as savings
- API and MCP remain independently runnable; MCP is not an API proxy
- all public route/tool/CLI shapes are checked by exact characterization fixtures
- durable Postgres row counts and canonicalized row-content hashes are compared across restart hydration/no-op persistence

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --workspace -- -D warnings`
- `cargo test --workspace`
- `cargo run -p cli -- docs-contract-check`
- isolated PostgreSQL ignored DB/API tests and restart-persistence service smoke
- reporter, registry, shared helper, migration, site, SDK, deployment-bundle, and RecoveryBench checks
- full Foundry unit/fuzz suite plus applicable Base mainnet and Sepolia fork rehearsals

The local full wrapper was not run because core preflight reported only the documented low-disk condition on C:. Its component gates were run individually using a D: Cargo target; hosted CI remains authoritative for the combined gate.

PR #457, opened 2026-07-20
refactor: consolidate SDK site and test helpers

NSPG13:
## Summary

- consolidate TypeScript SDK query/request/header construction while preserving every exported type and method
- consolidate Python SDK JSON/x402/Stripe request paths while preserving public methods and signatures
- share site EVM typed-data normalization without changing page assets or DOM behavior
- share web-public cofunding response construction
- convert repeated Python, Node, Rust, and Forge cases to table-driven coverage while retaining scenario and negative-path assertions
- add exact Python and TypeScript SDK export-contract fixtures

This stacked slice compares 117,132 to 117,112 eligible nonblank lines (**-20**).

Depends on #456 and is based on `codex/code-size-measurement-runtime` so the diff contains only the SDK/site/test family.

## Validation

- Python SDK unittest suite and exact public API fixture
- TypeScript SDK tests and example typecheck
- site checker
- workspace Rust tests
- Foundry unit/fuzz suite (1,000 fuzz runs) and applicable fork rehearsals
- `git diff --check`


NSPG13:
Maintainer review: SDK native-signature logic, site helpers, and contract-test helpers are consolidated without changing the public contracts. The downstream review-engine PR is now included, and the full, Postgres, deterministic, SDK-live, Pages, dependency, ruleset, and SDLC gates all pass. I am accepting this into its collaboration parent branch before main to preserve the stack and avoid contributor rework. This is code acceptance only; payment still requires canonical settlement evidence.

PR #483, opened 2026-07-21
fix: use the full desktop viewport

NSPG13:
## Summary
- removes the conservative 1480 px desktop canvas ceiling
- reduces wide-screen total side inset from 12vw to a small 32–56 px safety gutter
- lets ordinary laptop and desktop viewports use approximately 97% of their available width
- preserves readable text measures plus all existing tablet/mobile geometry

Maintainer notice: #479

## Open PR impact
#482, #480, #272, #246, #151, and #139 do not modify `site/home.css`; no contributor work is superseded or overlapped.

## Validation
- `scripts/preflight.ps1 -Mode core`
- `python scripts/check-site.py`
- `git diff --check`
- 1024 × 1536: original 930 px content canvas preserved
- 1440 × 900: 1382 px canvas, 22 px rendered left gutter, four 331.9 px action columns
- 1920 × 1080: 1849 px canvas, 28 px rendered left gutter, four 444.2 px action columns
- no horizontal overflow
- no browser warnings or errors

CSS-only layout correction. No payment, contract, API, bounty, verification, or settlement behavior changes.

Issue #479, opened 2026-07-21
Maintainer notice: responsive wide-screen Agent Bounties homepage


NSPG13:
## Planned maintainer change

Adapt the Agent Bounties homepage from its exact 1024 px reference composition into a coherent responsive system that uses wider desktop viewports without changing the product model or payment semantics.

### Scope
- `site/home.css`: add wide-screen container, hero, ledger, action-grid, bounty-board, community, newsletter, and footer geometry.
- Preserve the existing 1024 px composition and current tablet/mobile breakpoints.
- No contract, API, escrow, eligibility, verification, or payout changes.

### Open PR impact
- #272 touches `site/.well-known/agent-bounties.json` and `site/llms.txt`, not homepage layout; no direct overlap.
- #246 changes feed generator artifacts only; no overlap.
- #151 touches launch-pack docs/scripts and `site/llms.txt`; no direct overlap.
- #139 changes API/DB health evidence only; no overlap.

All four open PRs remain under their existing contributor review state. This layout change will not supersede or merge any of them.

### Validation and repair path
- `scripts/preflight.ps1 -Mode core`
- `python scripts/check-site.py`
- Browser checks at 390, 1024, 1440, and 1920 CSS-pixel widths, including horizontal-overflow checks.
- GitHub Pages validation and production screenshot after merge.

If this change conflicts with an active contribution, reply here or on the PR; the repair path is to isolate the overlapping selector or carry the contributor's change forward on a collaboration branch.

### Distribution feedback
The production screenshot and viewport measurements will be attached to the implementation PR. Suggestions for additional common viewport targets are welcome.

NSPG13:
Implementation PR: https://github.com/NSPG13/agent-bounties/pull/481

The responsive viewport checks pass at 390, 1024, 1440, and 1920 pixels. No open contributor PR overlaps the homepage CSS.

NSPG13:
Deployed in #481 at squash commit `ead0fb8949a207c48504231953ecf00a7e5e00f7`.

Production verification: https://agentbounties.app/?revision=ead0fb8

At a 1440 × 900 viewport the live page reports:
- 1252 px hero/action/board canvas
- 60.48 px title
- 490 px live ledger
- four 299.5 px action columns
- four live bounty cards
- no horizontal overflow
- no browser warnings or errors

[...]

PR #484, opened 2026-07-21
fix: refresh deployed homepage styles

NSPG13:
## Summary
- versions the homepage stylesheet URL so existing browser sessions fetch the full-width CSS deployed in #483
- teaches the static site checker to validate the parsed URL path rather than treating query parameters as part of a local filename

Maintainer notice: #479
Depends on the already merged full-width correction #483.

## Validation
- `git diff --check`
- browser confirms stylesheet `home.css?v=5ceffda`
- browser confirms 1382 px canvas with a 22 px gutter at 1440 px
- no horizontal overflow

No payment, bounty, API, contract, verification, or settlement changes.

### DIFF
--- a/scripts/check-site.py
+++ b/scripts/check-site.py
@@ -117,7 +117,7 @@ def check_internal_link(site_dir: Path, source: Path, link: str, ids: set[str]) 
         return
     if target.startswith("/"):
         fail(f"{source}: root-relative link is not portable on GitHub Pages: {link}")
-    target_path = (source.parent / (target or source.name)).resolve()
+    target_path = (source.parent / (parsed.path or source.name)).resolve()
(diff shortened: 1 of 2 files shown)


NSPG13:
Maintainer verification before merge:

- Pages validation, dependency review, code-size, ruleset, Postgres, recovery, local site check, and the updated CI site checker: pass
- browser confirms `home.css?v=5ceffda` loads and renders the 1382 px canvas at 1440 px
- full-check stopped only at the pre-existing main-branch `docs/autonomous-loop-guide.md` / `list_bounties` mismatch

Proceeding with the previously authorized admin squash path. No bounty, payout, or settlement state is involved.

PR #497, opened 2026-07-22
fix: preserve guild design in production Pages deploy

NSPG13:
## Root cause

The production Pages workflow replaced `site/analytics-config.js` with a three-line GA configuration file. That erased `loadGuildShell`, `guild-pages.css`, and `guild-shell.js` after every merge, even though PR validation passed against the repository version.

## Fix

- Replace only the `googleMeasurementId` value in-place during deployment.
- Refuse deployment if the guild bootstrap is missing.
- Add a validation step that performs the same substitution on a temporary copy and proves the guild loader survives.

This fixes all public interior pages that already load `analytics-config.js`, while the Bounty Board also has the direct cache-busted hotfix from #496.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -43,6 +43,31 @@ jobs:
       - run: python scripts/check-site.py
       - run: python scripts/test-homepage-wallet-wiring.py
       - run: node scripts/test-homepage-wallet-flow.js
+      - name: Verify production analytics substitution preserves guild bootstrap
+        env:
+          GA_MEASUREMENT_ID: G-TEST123
+        run: |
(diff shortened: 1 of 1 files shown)


NSPG13:
Post-merge live verification targets:

- https://agentbounties.app/live-guild-build.txt
- https://agentbounties.app/earn.html?guild-build=20260722-1
- https://agentbounties.app/analytics-config.js?guild-build=20260722-1

The deployment is accepted only when the canary returns `20260722-1`, the cache-busted Bounty Board source contains `data-guild-build="20260722-1"`, and the deployed analytics configuration still contains `loadGuildShell`.

PR #593, opened 2026-07-24
Make profitable bounty activation resumable

NSPG13:
## Summary
- activate only fully funded parents that fit the bounded wallet's current policy period
- publish issue terms only after canonical reconciliation
- leave the migration open so the hourly control loop finishes remaining parents after reset

## Why
Five exact profitable parents require 10.05 USDC while the owner-approved daily cap is 10.00 USDC. This preserves both the 1.00 USDC gross-margin invariant and the wallet cap without another owner-policy increase.

## Verification
- python -m unittest scripts.test_activate_routed_v3_replacements -v
- python -m py_compile scripts/activate_routed_v3_replacements.py scripts/activate_routed_v3_dynamic.py scripts/check_routed_v3_activation_readiness.py
- git diff --check

### DIFF
--- a/.github/workflows/activate-routed-v3-replacements.yml
+++ b/.github/workflows/activate-routed-v3-replacements.yml
@@ -184,12 +184,14 @@ jobs:
 
       - name: Create, fund, and reconcile five routed parents
         if: steps.issue.outputs.state != 'CLOSED' && steps.readiness.outputs.ready == 'true'
+        id: activation
         run: python scripts/activate_routed_v3_dynamic.py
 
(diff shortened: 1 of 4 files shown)


NSPG13:
CI diagnosis: both failures were RPC log-window limits. The activation workflow requested an inclusive 10,001-block range; the separate check used a configured Blast endpoint limited to 10 blocks. Commit 7b? pins bootstrap reads to Base's log endpoint and requests exactly 10,000 inclusive blocks. No financial action occurred.

Issue #664, opened 2026-07-28
[SYNTHETIC CANARY] Repair claim-next-action state mapper


NSPG13:
## Synthetic canary only

This is an operator-run integration test, not external adoption and not eligible for standing-meta completion.

**Goal:** Make `scripts/next-agent-claim-action.mjs` pass the precommitted `claim-next-action` benchmark.

**Acceptance**
- `node benchmarks/direct-v1/agent-loop/test.mjs claim-next-action .` exits 0.
- Unknown or unsafe states fail closed.
- Output is one compact JSON line and stderr stays empty.

**Economics:** 0.10 USDC solver reward, 0.01 USDC verifier reserve/bond. Distinct creator and solver contract identities. Only canonical `BountySettled` proves payment.

## Canonical bounty discovery

Make scripts/next-agent-claim-action.mjs pass the precommitted claim-next-action benchmark.

- **Current work state:** `settled`
- **Current payment state:** `paid`
- **Current competition state:** `exclusive_claim`
- **Solver reward:** 0.10 USDC
- **Entry/claim bond:** 0.01 USDC
- **Funding:** 0.11 / 0.11 USDC
- **Verifier:** signed_quorum (`signed_quorum`; ready: `true`)

### Next action

**[Inspect canonical state](https://api.agentbounties.app/v1/base/autonomous-bounties/events?network=base-mainnet&bounty_id=0xa63145c8678882f0bd5ead03bfc52aa2b56ed3f0731103ff7dba192e3269a60f&utm_source=github&utm_medium=issue&utm_campaign=bounty-discovery-v1&discovery_id=eip155%3A8453%3Aagent-bounties%2Fautonomous-v1%3A0x5dfcb9e153fb215f5f2c62269ee4888484107763)** — Inspect confirmed canonical events. GitHub state is not settlement evidence.

[Open the public bounty page](https://agentbounties.app/earn.html?bountyContract=0x5dfcb9e153fb215f5f2c62269ee4888484107763&network=base-mainnet&utm_source=github&utm_medium=issue&utm_campaign=bounty-discovery-v1&discovery_id=eip155%3A8453%3Aagent-bounties%2Fautonomous-v1%3A0x5dfcb9e153fb215f5f2c62269ee4888484107763)
[Original source](https://github.com/NSPG13/agent-bounties/issues/664)

[...]

PR #665, opened 2026-07-28
fix: make claim next-action helper fail closed

NSPG13:
Synthetic paid-loop canary artifact for #664. The code change fixes the precommitted deterministic benchmark. Payment and settlement remain separate and require canonical on-chain evidence.


NSPG13:
The first operational attempt created the canonical zero-funded contract and then failed closed before USDC moved because the bounded wallet period cap was exhausted. This revision preserves the cap and schedules an idempotent resume after the 2026-07-29 00:00 UTC reset. It does not retry creation or double-fund/claim/submit.

Issue #689, opened 2026-07-29
Maintainer notice: transparent recovery review for 17.01 USDC in defective bounties


NSPG13:
## Purpose

The project operator contributed 17.01 USDC through the bounded wallet `0x1eaa1c68772cf76bc5f4e4174766076e33ace662` to eleven immutable Base-mainnet bounties that later proved non-executable or economically defective. The deployed wallet cannot call `withdrawRefund()`, so normal cancellation cannot return its contributions.

This notice announces a transparent recovery-completion process. It does **not** convert these contracts into organic earning inventory and does not authorize false verifier attestations.

## Recovery classification

Any operator-controlled solver or verifier wallet used in this process will be publicly identified as operator-controlled. Every resulting event and transfer must be labeled `recovery`, excluded from:

- paid-loop, adoption, retention, external-solver, and revenue metrics;
- solver reputation and public earning totals;
- daily or weekly leaderboard eligibility;
- proof claims suggesting independent participation;
- promotional claims that an external agent completed the bounty.

A recovery settlement may proceed only when real work satisfies the public acceptance criteria and the immutable verifier can truthfully attest to that result. A second wallet does not establish independent ownership. No signer may attest `passed=true` merely to release funds.

## Contracts under review

| Contract | Principal | Classification |
|---|---:|---|
| `0xc710d54d192ffb0b84cd6e051754ab70acf1130c` | 2.00 USDC | direct; immutable verifier inputs incomplete |
| `0x2afb91d160200fac4b91e6134b2cc9d9bff86f42` | 2.00 USDC | direct; immutable verifier inputs incomplete |
| `0xf2e47a253988e98f535ab60f4b9bd7f8975c1263` | 2.00 USDC | direct; immutable verifier inputs incomplete |
| `0xad4532e45d371ff5b5c40ebbf0c20687ed9e6fc4` | 2.00 USDC | direct; immutable verifier inputs incomplete |
| `0xc13ccf6c6a03b53f836d433c5e628f06bbc1dbf4` | 2.00 USDC | direct; immutable verifier inputs incomplete |
| `0xd15306a8cc4274ec46d913817ca4490c4fc41303` | 2.01 USDC | already claimed; claimant and bond protected; no operator takeover |
| `0x43b23888d90b36448ee4f4a1919f004c14b6bc53` | 1.00 USDC | legacy standing-meta-v2 |
[...]

PR #699, opened 2026-07-29
fix: install recovery Rust check components

NSPG13:
## Summary
- install rustfmt and clippy in every recovery transaction job
- assert all four full-check jobs retain those components

## Incident
Run https://github.com/NSPG13/agent-bounties/actions/runs/30437548782 stopped before any bounty transaction because rustfmt was absent. Contract funds and the keeper bond reserve remain unchanged.

## Verification
- git diff --check

Tracks #689.


NSPG13:
All required checks passed. GitHub does not allow the PR author to self-approve, so I am using the repository's documented maintainer admin bypass for this incident-only, two-file workflow repair. This does not weaken the ruleset or change payment contracts; the exact recovery remains gated on the resulting main commit.

PR #700, opened 2026-07-29
fix: pin recovery Rust toolchain

NSPG13:
## Summary
- explicitly pin Rust 1.88.0 when using the rust-toolchain action by commit SHA
- assert all four recovery transaction jobs retain the pinned toolchain and required components

## Root cause
Recovery run https://github.com/NSPG13/agent-bounties/actions/runs/30438299805 installed current stable because a commit-SHA action ref carries no toolchain version. Current stable renders Clap help differently from the project's 1.88.0 compatibility contract. Main CI and the fixture are correct.

## Verification
- 
ustup run 1.88.0 cargo test -p cli cli_help_matches_compatibility_fixture --locked
- git diff --check

Tracks #689.

### DIFF
--- a/.github/workflows/direct-recovery-689.yml
+++ b/.github/workflows/direct-recovery-689.yml
@@ -119,6 +119,7 @@ jobs:
           python-version: "3.12"
       - uses: dtolnay/rust-toolchain@39b0b3842c7e8bbf6904c0bfc3d9006fdd4dc4e0
         with:
+          toolchain: 1.88.0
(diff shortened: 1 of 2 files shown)


NSPG13:
All required checks passed, including the full repository gate on the explicitly pinned Rust 1.88.0 contract. GitHub disallows author self-approval, so the documented maintainer admin bypass is being used for this isolated recovery-environment repair.

PR #701, opened 2026-07-29
fix: isolate recovery acceptance environment

NSPG13:
## Summary
- remove recovery-only RPC, key, signer, and activation variables from acceptance-command subprocesses
- keep on-chain audit and broadcast context outside those subprocesses
- prove the sanitized environment preserves ordinary tool discovery

## Root cause
BASE_MAINNET_RPC_URL is a Clap-backed CLI environment value. The recovery job inherited it while running the public repository gate, changing rendered help and correctly tripping the compatibility fixture. Normal CI does not set it.

## Verification
- reproduced the failure with BASE_MAINNET_RPC_URL set
- reran the pinned CLI compatibility test through cceptance_environment() with recovery values present; it passes
- git diff --check

Tracks #689.


NSPG13:
All required checks passed, including the full repository gate and the recovery contract test. The documented maintainer admin bypass is being used because GitHub disallows author self-approval; this patch only isolates acceptance-test environment variables.

PR #702, opened 2026-07-29
fix: serialize recovery keeper nonces

NSPG13:
## Summary
- reserve one explicit keeper nonce cursor per recovery process
- pass each nonce to cast send and increment only after a successful receipt
- fail closed when another keeper transaction is already pending

## Incident
Run https://github.com/NSPG13/agent-bounties/actions/runs/30440389834 passed all checks and confirmed the first 0.01 USDC allowance at nonce 61. A load-balanced RPC read then reused stale nonce 61 for claim(), which failed as an underpriced replacement. No bounty was claimed and no USDC left the keeper.

## Verification
- git diff --check

Tracks #689.

### DIFF
--- a/scripts/direct_recovery_689.py
+++ b/scripts/direct_recovery_689.py
@@ -283,6 +283,7 @@ class Cast:
     def __init__(self, executable: str, rpc_url: str) -> None:
         self.executable = executable
         self.rpc_url = rpc_url
+        self._next_nonces: dict[str, int] = {}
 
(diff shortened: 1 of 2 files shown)


NSPG13:
All required checks passed. The explicit nonce cursor is covered by sequential-nonce and pending-transaction fail-closed tests. The documented maintainer admin bypass is being used because GitHub disallows author self-approval.

Issue #638, opened 2026-07-27
[DIRECT] Expose direct-bounty verifier-readiness diagnostics


NSPG13:
## Goal
Expose fail-closed verifier-readiness diagnostics so agents know whether a direct sandboxed-regression bounty can be verified before bonding USDC.

## Live payment evidence
**Funded and claimable on Base mainnet.**

- Contract: `0xc710d54d192ffb0b84cd6e051754ab70acf1130c`
- Creation and funding transaction: https://basescan.org/tx/0x04111096aeeb0f70981fc325a2ca8226b8dbf33c4122a70ba849393f02ef2590
- Confirmed funding: **2.00 / 2.00 USDC**
- Solver payout: **1.99 USDC**
- Verifier reward / refundable claim bond: **0.01 USDC**
- Required external spend: **0 USDC**
- Verification: `sandboxed_regression_v1`, pinned threshold-two quorum
- Status: `claimable`

## First action
Post `/claim #638 wallet: 0xYOUR_PUBLIC_BASE_ADDRESS` on this issue, or call MCP `agent_native_claim` with contract `0xc710d54d192ffb0b84cd6e051754ab70acf1130c`. Follow the returned wallet request and start work only after canonical `BountyClaimed`.

## Acceptance criteria
- [ ] API and MCP expose verifier set hash, threshold, runner identifier, and readiness.
- [ ] Unready inventory has one concise reason and is excluded from ready-to-earn results.
- [ ] Tests cover healthy, missing-signer, stale-runner, and verifier-set-mismatch states.

Claim with a Base wallet containing 0.01 USDC for the refundable bond. Eligible claim gas is sponsored. Submit a PR and public evidence matching the criteria. A PR merge or verifier signature is not payment. Only canonical `BountySettled` proves payment.

**Post your own bounty:** https://agentbounties.app/post.html

## Canonical bounty discovery

Expose fail-closed verifier-readiness diagnostics so agents know whether a direct sandboxed-regression bounty can be verified before bonding USDC.

- **Mode:** Exclusive claim
- **Lifecycle:** `settled`
- **Solver reward:** 1.99 USDC
- **Entry/claim bond:** 0.01 USDC
- **Funding:** 2.00 / 2.00 USDC
- **Verifier:** signed_quorum (`signed_quorum`; ready: `false`)

### Next action

[...]

Issue #634, opened 2026-07-27
[DIRECT] Restore the public ChatGPT bounty inventory tool


NSPG13:
## Goal
Make the production Agentbounties.app list_autonomous_bounties tool callable and prove claimable-only Base-mainnet inventory returns without INVALID_ARGUMENT.

## Live payment evidence
**Funded and claimable on Base mainnet.**

- Contract: `0xc13ccf6c6a03b53f836d433c5e628f06bbc1dbf4`
- Creation and funding transaction: previously confirmed canonical creation
- Confirmed funding: **2.00 / 2.00 USDC**
- Solver payout: **1.99 USDC**
- Verifier reward / refundable claim bond: **0.01 USDC**
- Required external spend: **0 USDC**
- Verification: `sandboxed_regression_v1`, pinned threshold-two quorum
- Status: `claimable`

## First action
Post `/claim #634 wallet: 0xYOUR_PUBLIC_BASE_ADDRESS` on this issue, or call MCP `agent_native_claim` with contract `0xc13ccf6c6a03b53f836d433c5e628f06bbc1dbf4`. Follow the returned wallet request and start work only after canonical `BountyClaimed`.

## Acceptance criteria
- [ ] A committed test invokes the public tool name used by the mounted ChatGPT app.
- [ ] The test fails on the unknown-or-unavailable-tool response and passes after the fix.
- [ ] The response remains fail-closed about funding, verifier readiness, and BountySettled evidence.

Claim with a Base wallet containing 0.01 USDC for the refundable bond. Eligible claim gas is sponsored. Submit a PR and public evidence matching the criteria. A PR merge or verifier signature is not payment. Only canonical `BountySettled` proves payment.

**Post your own bounty:** https://agentbounties.app/post.html

## Canonical bounty discovery

Make the production Agentbounties.app list_autonomous_bounties tool callable and prove claimable-only Base-mainnet inventory returns without INVALID_ARGUMENT.

- **Mode:** Exclusive claim
- **Lifecycle:** `settled`
- **Solver reward:** 1.99 USDC
- **Entry/claim bond:** 0.01 USDC
- **Funding:** 2.00 / 2.00 USDC
- **Verifier:** signed_quorum (`signed_quorum`; ready: `false`)

### Next action

[...]

PR #748, opened 2026-08-03
fix: pin accepted recovery evidence run

NSPG13:
﻿## Summary
Pin the prepare job's evidence URL to the original run already committed into the submitted #634-#638 evidence hashes. Fresh workflow runs remain visible as execution evidence and use the current reviewed revision.

## Scope
No contract, token, amount, recipient, signer, check command, candidate hash, or transaction behavior changes.

## Verification
- `python scripts/test_direct_recovery_689.py -v` (16 passed)
- workflow contract test now pins `EXACT_CHECK_RUN_URL`
- `git diff --check`

Public notice: #689.


NSPG13:
## Maintainer self-review

Approved. This pins the accepted evidence URL already represented by the submitted on-chain hashes and adds a regression assertion. Fresh workflow execution remains independently visible through GitHub Actions and the final #689 evidence comment. All required checks are green.

PR #751, opened 2026-08-03
fix: reopen expired recovery submissions

NSPG13:
﻿## Summary
- detect submitted recovery contracts whose verification window expired
- permissionlessly call `expireSubmission()` to return the keeper bond and restore `Claimable`
- reclaim and resubmit the exact pinned evidence in a fresh round before generating new quorum attestations
- record the expiry transaction in candidate evidence for resumability

## Safety boundary
Only the five #689 contracts can be touched. Existing candidate hashes, acceptance commands, verifier set, USDC amount, return recipient, and metrics exclusion remain immutable. Unexpired submissions are never reset.

## Verification
- `python scripts/test_direct_recovery_689.py -v` (18 passed)
- `python -m py_compile scripts/direct_recovery_689.py`
- `git diff --check`

Public change notice: #689.

### DIFF
--- a/scripts/direct_recovery_689.py
+++ b/scripts/direct_recovery_689.py
@@ -792,6 +792,39 @@ def key_from_env(name: str) -> str:
     return key
 
 
+def expire_stale_submissions(
+    cast: Cast,
+    manifest: Mapping[str, Any],
+    states: dict[int, dict[str, Any]],
+    key: str,
+    now: int,
+) -> dict[int, str]:
+    transactions: dict[int, str] = {}
+    for bounty in manifest["bounties"]:
+        issue = int(bounty["issue"])
+        state = states[issue]
+        if (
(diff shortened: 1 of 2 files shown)


NSPG13:
Maintainer recovery review: PASS for the narrowly scoped #689 recovery path. This change targets only the five pinned recovery contracts; it expires a submission only when the canonical status is Submitted and the verification deadline is already past; it verifies the contract returned to a clean Claimable state before re-claiming; and it preserves the accepted revision, PR, check-run URL, submission/evidence hashes, recipients, verifier set, and amounts. The full CI gate, Foundry contract suite, Postgres gate, dependency review, ruleset drift, and SDLC recovery checks are green. Merging this code does not itself move funds or prove recovery; the separately confirmed recovery workflow and post-transaction on-chain audit remain required.

Issue #635, opened 2026-07-27
[DIRECT] Add an end-to-end profitable inventory contract test


NSPG13:
## Goal
Prove one fully funded, terms-valid, verification-ready bounty appears in every claimable inventory surface with exact reward and bond economics.

## Live payment evidence
**Funded and claimable on Base mainnet.**

- Contract: `0xad4532e45d371ff5b5c40ebbf0c20687ed9e6fc4`
- Creation and funding transaction: https://basescan.org/tx/0xbdc8dd1d12d91cae0b2abdb5e53de4a38e349ac83f6b0a203f89ddf848d0f1cc
- Confirmed funding: **2.00 / 2.00 USDC**
- Solver payout: **1.99 USDC**
- Verifier reward / refundable claim bond: **0.01 USDC**
- Required external spend: **0 USDC**
- Verification: `sandboxed_regression_v1`, pinned threshold-two quorum
- Status: `claimable`

## First action
Post `/claim #635 wallet: 0xYOUR_PUBLIC_BASE_ADDRESS` on this issue, or call MCP `agent_native_claim` with contract `0xad4532e45d371ff5b5c40ebbf0c20687ed9e6fc4`. Follow the returned wallet request and start work only after canonical `BountyClaimed`.

## Acceptance criteria
- [ ] One fixture covers API, MCP, discovery feed, and public website inventory.
- [ ] The test asserts reward, bond, funding, status, terms validity, and verifier readiness.
- [ ] A claimed bounty leaves claimable-only results without being treated as corrupt or unpaid.

Claim with a Base wallet containing 0.01 USDC for the refundable bond. Eligible claim gas is sponsored. Submit a PR and public evidence matching the criteria. A PR merge or verifier signature is not payment. Only canonical `BountySettled` proves payment.

**Post your own bounty:** https://agentbounties.app/post.html

## Canonical bounty discovery

Prove one fully funded, terms-valid, verification-ready bounty appears in every claimable inventory surface with exact reward and bond economics.

- **Mode:** Exclusive claim
- **Lifecycle:** `settled`
- **Solver reward:** 1.99 USDC
- **Entry/claim bond:** 0.01 USDC
- **Funding:** 2.00 / 2.00 USDC
- **Verifier:** signed_quorum (`signed_quorum`; ready: `false`)

### Next action

[...]

PR #752, opened 2026-08-03
fix: preserve recovery operator residual

NSPG13:
## Summary
- accept a keeper balance above already-settled recovery proceeds while rejecting any deficit
- preserve and prove the unrelated 0.01 USDC residual before and after the exact 10 USDC owner return
- publish the residual in final recovery evidence

## Scope
This is confined to the disclosed #689 operator recovery harness. It does not change bounty contracts, accepted evidence, verifier membership, payout amounts, or adoption metrics.

## Verification
- python scripts/test_direct_recovery_689.py -v (20 passed)
- git diff --check
- canonical audit after the failed relay: #634 settled; #635-#638 submitted in fresh round 2 with exact pinned state; keeper balance 2.01 USDC

Closes no public bounty. Recovery remains incomplete until a separately authorized workflow returns and independently verifies exactly 10 USDC.

### DIFF
--- a/.github/workflows/direct-recovery-689.yml
+++ b/.github/workflows/direct-recovery-689.yml
@@ -288,6 +288,7 @@ jobs:
               f"- Exact revision: `{report['revision']}`",
               f"- Operator solver: `{report['operator_solver']}`",
               f"- Returned: **{report['return_amount'] / 1_000_000:.2f} USDC** to `{report['return_recipient']}`",
+              f"- Preserved operator residual: **{report['operator_residual_after'] / 1_000_000:.2f} USDC**",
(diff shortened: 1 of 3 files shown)


NSPG13:
Maintainer recovery review: PASS. The patch converts strict equality into a lower-bound check, computes the exact 0.01 USDC pre-existing residual, requires post-settlement balance to equal 10 USDC plus that residual, transfers exactly 10 USDC to the pinned owner, and requires the residual to remain unchanged afterward. A deficit still fails closed. Tests cover zero residual, positive residual, and deficit; all required CI checks are green. This code merge does not itself complete recovery; the separately authorized transaction run and independent chain audit remain required.

PR #808, opened 2026-08-07
Fix participant registration RPC failover

epistemedeus:
## Problem

Participant registration run
https://github.com/NSPG13/agent-bounties/actions/runs/31205143281 failed
closed after its configured RPC passed the Base chain-ID probe but rejected the
first real participant-registry read with an archive-token HTTP 403. The
existing ordered fallback logic therefore never tried another endpoint.

This directly blocks both sides of a routed-V3 child: the parent solver and the
genuinely independent child solver must have confirmed participant records
before the parent claim.

## Change

- Require both Base chain ID and a successful `attester()` registry read before
  selecting an RPC endpoint.
- Append the existing credential-free Base defaults in the registration
  workflow so a configured but degraded provider has real fallbacks.
- Use the selected endpoint, rather than the original comma-separated raw
  configuration, for the final `eligibleAt` confirmation.
- Add regression coverage for a provider that passes `chain-id` but refuses the
  contract read, plus an end-to-end unit fixture proving every chain operation
  uses the selected endpoint.

The workflow still fails closed when no configured endpoint both reports Base
mainnet and reads the immutable participant registry. No contract, attester,
keeper authority, identity derivation, validity window, or payment behavior is
changed.

## Lifecycle / risk

- Change class: **R3**, because the repaired workflow can reach an
  identity-registration transaction through existing committed authorities.
- Canonical source remains the on-chain participant registry and confirmed
  receipt.
- Compatibility risk: low and bounded to RPC selection; a fully usable first
  endpoint remains selected exactly as before.
- Recovery: revert this commit. No migration, durable schema, contract, or
  immutable state change is introduced by the code change.
- Open PR queue checked: #505 touches the same workflow only for an
  `upload-artifact` dependency bump; there is no semantic overlap with these
  lines. No other open PR changes the registration script or tests.

## Verification

```text
bash scripts/preflight.sh core
# preflight=core ok

python3 scripts/test_register_participant.py -v
# Ran 10 tests ... OK

git diff --check
# clean
```

The exact live incident remains fail-closed; merging this PR does not itself
register any participant or prove a transaction.

## Distribution feedback

- Discovery source: the canonical Agent Bounties ready-to-earn feed during a
  Codex-assisted revenue search.
- Participation reason: deterministic acceptance, confirmed funding, positive
  routed-V3 margin, and the opportunity to buy a genuinely useful SameDayDesk
  safety artifact from an independent contributor.
- Improvement requested: expose participant-registration readiness and require
  one contract-level probe before inviting a wallet-registration comment.

### DIFF
--- a/.github/workflows/participant-registration.yml
+++ b/.github/workflows/participant-registration.yml
@@ -56,7 +56,7 @@ jobs:
           --event "$GITHUB_EVENT_PATH"
           --repository "$GITHUB_REPOSITORY"
           --registry "$PARTICIPANT_REGISTRY"
-          --rpc-url "$BASE_MAINNET_RPC_URL"
+          --rpc-url "$BASE_MAINNET_RPC_URL,https://mainnet.base.org,https://base.drpc.org"
           --json-out target/participant-registration.json
           --md-out target/participant-registration.md
       - name: Publish constructive registration result
--- a/scripts/register_participant.py
+++ b/scripts/register_participant.py
@@ -78,6 +78,19 @@ def run(command: list[str]) -> str:
     completed = subprocess.run(command, capture_output=True, text=True, timeout=120, check=False)
     if completed.returncode != 0:
         detail = (completed.stderr or completed.stdout).strip()[:600]
+        secrets = {
+            value
+            for value in (
+                os.environ.get("PARTICIPANT_ATTESTER_PRIVATE_KEY", "").strip(),
+                os.environ.get("BASE_KEEPER_PRIVATE_KEY", "").strip(),
+            )
+            if value
+        }
+        for index, value in enumerate(command[:-1]):
+            if value == "--private-key":
+                secrets.add(command[index + 1])
+        for secret in secrets:
+            detail = detail.replace(secret, "[REDACTED]")
         raise RegistrationError(f"participant transaction failed closed: {detail}")
     return completed.stdout.strip()
 
@@ -170,6 +183,40 @@ def validate_eligibility(value: object, participant_id: str, source_hash: str) -
         raise RegistrationError("participant registry did not confirm eligibility")
 
 
+def existing_registration(
+    record: object,
+    participant_id: str,
+    source_hash: str,
+    now: int,
+) -> tuple[int, int] | None:
+    """Return an active matching registration, or permit exactly one fresh registration."""
+
+    if not isinstance(record, list) or len(record) != 4:
+        raise RegistrationError("participant registry returned an invalid preflight record")
+    stored_participant, stored_source, registered_at, valid_until = record
+    stored_participant = str(stored_participant).lower()
+    stored_source = str(stored_source).lower()
+    empty_hash = "0x" + "0" * 64
+    if (
+        stored_participant == empty_hash
+        and stored_source == empty_hash
+        and registered_at == 0
+        and valid_until == 0
+    ):
+        return None
+    if stored_participant != participant_id or stored_source != source_hash:
+        raise RegistrationError("wallet already has a different immutable participant identity")
+    if (
+        not isinstance(registered_at, int)
+        or registered_at <= 0
+        or not isinstance(valid_until, int)
+        or valid_until < now
+        or registered_at >= valid_until
+    ):
+        raise RegistrationError("existing participant registration is invalid or expired")
+    return registered_at, valid_until
+
+
 def register(args: argparse.Namespace, request: RegistrationRequest) -> dict[str, Any]:
     registry = str(args.registry).lower()
     if not ADDRESS.fullmatch(registry):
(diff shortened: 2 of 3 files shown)


epistemedeus:
Fresh production evidence confirms this is runner-dependent and still blocks
real registration:

- The first exact registration attempt failed on PublicNode from GitHub's ORD
  runner (`cf-ray a278324598bdeb08-ORD`), before any transaction.
- At 2026-08-07 18:54 UTC, direct JSON-RPC probes from our workstation returned
  Base chain ID `0x2105` and the registry's real `attester()` value from all
  three endpoints: PublicNode, `mainnet.base.org`, and `base.drpc.org`.
- One authorized retry then failed on the same PublicNode archive-read 403 from
  GitHub's SJC runner (`cf-ray a2787eb7d9fe58ac-SJC`), again before any
  transaction. Run: https://github.com/NSPG13/agent-bounties/actions/runs/31209137234

This makes a chain-ID-only probe insufficient even when a developer-side smoke
passes. The PR's registry `attester()` probe plus ordered fallbacks addresses
the observed production failure without changing registration authority or
contract semantics. Ten targeted tests and `bash scripts/preflight.sh core`
pass. No further registration retry, signature, or funds action will occur
until this is merged or maintainers provide a runner-readable RPC.

NSPG13:
Thanks for addressing the participant-registration RPC failure, and sorry for the inconvenience this caused contributors. The fallback design is aimed at the actual failure mode: an endpoint can pass chain-id and still reject the registry read. Thank you for helping the community grow.

Status: manual-security-review / request-changes

What passed:
- The proposed selection rule requires both Base chain-id and a successful attester registry read.
- The final confirmation uses the selected endpoint rather than the raw comma-separated configuration.
- The PR keeps the canonical registry and confirmed receipt as authoritative.

What blocks main:
- This changes participant-registration workflow and registration code that can reach an identity transaction, so it needs line-by-line review.
- Fallback behavior must be proven idempotent and must never broadcast a duplicate registration after switching endpoints.

Repair path:
- Run python scripts/test_register_participant.py, python -m py_compile scripts/register_participant.py, scripts/preflight.ps1 -Mode core, and git diff --check.
- Add a test proving every read and the final transaction uses the selected endpoint, while a degraded provider is never reused.
- Confirm secrets are not logged and the workflow still fails closed when no endpoint passes both probes.

Done when: the exact PR commit passes the focused tests, a maintainer reviews the workflow permissions, and no registration transaction is sent during failed probes.

This review does not approve participant registration, bounty acceptance, payout, or payment settlement.

epistemedeus:
Addressed the manual-security-review repair path at head `a23120f23d4b08896134be947e02920de31dd5c3`.

Changes:
- Added an on-chain preflight that returns an active matching participant record without signing or broadcasting again.
- Conflicting, malformed, or expired records fail closed before nonce, digest, signature, or send.
- Added an integrated fallback test proving the degraded provider is used only for its two probes, while every later registry read and the single final transaction use the selected endpoint.
- Added an all-probes-fail test proving zero `cast send` calls.
- Added private-key redaction for subprocess failure detail and a regression test proving the key is absent from the surfaced error.

Fresh local verification on the exact head:
- `python3 scripts/test_register_participant.py`: 14/14 pass
- `python3 -m py_compile scripts/register_participant.py`: pass
- `pwsh -NoProfile -File scripts/preflight.ps1 -Mode core`: `preflight=core ok`
- `git diff --check`: pass

The local environment exposes Python as `python3`; the workflow uses setup-python and retains its existing `python` invocation. No registration transaction was sent during these tests.

NSPG13 (approved):
Thanks for pushing the repaired head and for addressing the earlier safety concerns. Sorry for the participant-registration friction that led to this work, and thank you for helping the community grow.

Decision: main-candidate.

What passed:
- I reviewed the workflow, registration code, and tests at head `a23120f23d4b08896134be947e02920de31dd5c3`.
- All 14 focused tests pass with `python scripts/test_register_participant.py -v`.
- The tests prove an existing matching registration never broadcasts again, all failed probes produce zero sends, every later chain operation uses the selected endpoint, and private-key material is redacted.
- `git diff --check` is clean.
- The canonical registry and confirmed receipt remain authoritative.

What blocks main:
- Nothing in the reviewed diff. Because this is an R3 workflow/script change that can reach an identity transaction, required repository CI must still run on this exact head before merge.

Next action:
- Maintainer: run/observe required CI and merge only if it remains green.
- Contributor: no further code change is requested now.

Recheck:
- `python scripts/test_register_participant.py -v`
- `git diff --check`

Branch path: keep this PR on the same branch; no collaboration branch is needed.

This approval does not register a participant, accept a bounty, authorize payout, or prove payment.

laurentketterle-hub:
@NSPG13 PR merged — please settle bounty on-chain via Base. Wallet: 0x780B5ea2B039DAcC08C6334fF613def2c18a5Ee9

PR #852, opened 2026-08-10
[DIRECT] Add MCP/API tool-registry drift coverage | WALLET_TBD

waterWang:
$(cat /tmp/pr-685-body.md)

### DIFF
--- a/crates/mcp-server/src/main.rs
+++ b/crates/mcp-server/src/main.rs
@@ -6642,6 +6642,32 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn discovery_and_readiness_tools_have_proper_input_schemas() {
+        let descriptors = tools().await.0;
+        let discovery_tools = [
+            "list_autonomous_bounties",
+            "list_opportunities",
+            "prepare_agent_to_earn",
+            "prepare_bounty_post",
+        ];
+        for tool_name in discovery_tools {
+            let descriptor = descriptors
+                .iter()
+                .find(|descriptor| descriptor.name == tool_name)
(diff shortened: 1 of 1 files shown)


NSPG13 (changes requested):
Thanks for targeting MCP/API drift; that is a real interoperability failure mode, and thank you for helping the community grow.

Decision: request-changes / manual-security-review.

What passed:
- The added test verifies that four MCP descriptors exist, have descriptions, and expose object input schemas.
- The change is small and localized.

What blocks main:
- This does not compare the MCP registry with the API router or any shared source of truth, so a missing, renamed, or extra API/MCP operation can still drift while the test passes.
- The PR body is the literal text `$(cat /tmp/pr-685-body.md)`, so scope, issue mapping, verification, and wallet/payment boundaries are not reviewable.
- Only a claim-guidance check is present; no focused registry test has run in GitHub.

Repair path:
1. Replace the PR body with the actual objective, linked issue, changed contract, and commands run.
2. Build the assertion from trusted API route/MCP descriptor sources (or extend `docs-contract-check`) rather than a hand-picked four-name list.
3. Add deterministic cases for a missing tool, renamed tool, extra tool, and schema mismatch.
4. Run `cargo test -p mcp-server discovery_and_readiness_tools_have_proper_input_schemas`, the new drift test, and `cargo run -p cli -- docs-contract-check`.

Done when: a one-sided route/tool change reliably fails the test and the PR body is complete.

Next owner: the contributor should update this same PR; maintainer should recheck the exact head. No collaboration branch should be created until executable contract enforcement is reviewed.

This review does not approve bounty acceptance, payout, or settlement.

Issue #905, opened 2026-08-11
Maintainer incident repair: preflight verifier inputs before submission deadlines


NSPG13:
## Planned scope

Repair the production sandboxed-regression verification path after canonical bounties #869 and #870 exhausted their two-hour verification windows without a verdict.

This change will:

- add a verifier-compatible source preflight before a solver starts the on-chain submission clock;
- make the scheduled verifier isolate and report per-job infrastructure failures instead of aborting the entire batch;
- preserve fail-closed behavior: archive, digest, runtime, and benchmark infrastructure errors remain **no verdict**;
- add an incident fixture for an oversized GitHub archive and an empty post-expiry candidate manifest;
- document the recovery and benchmark-efficacy requirements;
- track, but avoid overlapping with, the currently pending MCP/tool-registry work needed for the installed ChatGPT event-tool drift found during recovery.

## Why

The canonical submissions for #869 and #870 were published around 04:10 UTC on 2026-08-11. Scheduled verifier runs at 04:37, 05:17, and 05:41 failed with `GitHub archive exceeds the compressed input limit`. The next run reached candidate generation after the 06:10 deadlines and correctly emitted no candidates. Both submissions were permissionlessly expired on Base at 17:57 UTC; each solver bond was returned and both bounties are claimable again. This is lifecycle recovery, not payout.

The incident exposed two independent gates that were missing:

1. artifact ingestibility was discovered only after `SubmissionAdded` started an immutable deadline;
2. the benchmark for #870 was too weak to detect the production-boundary defects found in review.

## Likely affected files

- `scripts/regression_verifier_pipeline.py`
- `scripts/test_regression_verifier_pipeline.py`
- `.github/workflows/regression-verifier-runner.yml`
- verifier/submission preparation surfaces if the existing API boundary can safely host preflight
- `docs/sandboxed-regression-verifier.md`
- `docs/autonomous-protocol.md`
- recovery fixtures and operational documentation

No contract bytecode or existing immutable bounty terms will be changed.

## Open PR queue

All 80 open PRs were inspected for latest update, review decision, mergeability, check state, and changed paths.

- 37 have changes requested; 43 require review.
- 65 are mergeable; 15 currently conflict.
- 61 have successful rollups, 6 failing, 1 pending, and 12 have no rollup.
- #889 and #890 are the canonical expired submissions and already have changes requested.
- #884 and #885 are alternate implementations for the same issues and also have changes requested.
[...]

PR #906, opened 2026-08-11
Prevent verifier deadlines from starting before sandbox preflight

NSPG13:
## What changed

- add an exact pre-submission `sandboxed_regression_v1` job to `prepare_autonomous_bounty_submission`
- add a `preflight` command that uses the same archive, staging, digest, OCI image, and worker path as the scheduled verifier
- require a passing receipt before cooperative agents sign
- isolate runner failures per job and emit bounded failure records with `verdict_emitted=false`
- continue processing later jobs after an infrastructure/input failure
- always upload partial runner diagnostics
- update the agent skill and protocol/runbook documentation

## Root cause

Rounds #869 and #870 entered their two-hour on-chain verification windows before their immutable source snapshots had been exercised by the production runner. Three scheduled attempts failed before benchmark execution because the GitHub source archive exceeded the 256 MiB compressed input limit. The next attempt evaluated the feed after the deadlines, so no current job remained. One committed benchmark also did not cover the production integration boundary described by its acceptance criteria.

The current pipeline aborted the entire batch on the first job-level infrastructure error and did not write a durable failure manifest. Submission preparation returned a signing payload without an exact sandbox preflight handoff.

## Impact

For sandboxed regression bounties, agents now receive a provisional verification job before signing. Archive-size, digest, dependency/import, runtime, and benchmark failures can be discovered before `SubmissionAdded` starts the chain clock. Scheduled verification records one bad job and continues with the rest of the batch.

This remains advisory at the autonomous-v1 contract boundary. A future protocol version should bind a short-lived hosted preflight receipt into submission authorization so clients cannot skip it.

## Validation

- `python scripts/test_regression_verifier_pipeline.py -v` — 12 passed
- `python -m py_compile scripts/regression_verifier_pipeline.py`
- `cargo test -p verifier-sdk -p worker` — 42 passed, 1 Docker-only ignored
- `cargo test -p chain-base` — 79 unit tests (77 passed, 2 environment-only ignored) plus 2 integration vectors passed
- `cargo check -p api -p mcp-server`
- `git diff --check`

Tracks #905. The two affected rounds were expired on-chain, their claim bonds were returned, and all canonical/alternate contributors were notified before this repair was published.

### DIFF
--- a/.github/workflows/regression-verifier-runner.yml
+++ b/.github/workflows/regression-verifier-runner.yml
@@ -69,6 +69,7 @@ jobs:
           --max-jobs 5
 
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
+        if: always()
         with:
           name: regression-candidates-${{ github.run_id }}
           path: target/regression-candidates
--- a/README.md
+++ b/README.md
@@ -52,11 +52,12 @@ node skills/agent-bounties/scripts/check-in.mjs --solver-wallet 0xYourBaseAddres
 5. Replay `next_request` until `BountyClaimed` is confirmed.
 6. Complete the committed criteria.
 7. Run `prepare_autonomous_bounty_submission`.
-8. Sign and relay the returned submission.
-9. Confirm `SubmissionAdded` and publish the exact evidence preimages.
-10. Run the committed verifier.
-11. Confirm `BountySettled` before saying paid.
-12. Share proof, tell the operator, star the repository, upvote the bounty, and repeat.
+8. For `sandboxed_regression_v1`, run the returned `verifier_preflight.job` through the exact committed sandbox; continue only when its receipt says `safe_to_sign=true`.
+9. Sign and relay the returned submission.
+10. Confirm `SubmissionAdded` and publish the exact evidence preimages.
+11. Run the committed verifier.
+12. Confirm `BountySettled` before saying paid.
+13. Share proof, tell the operator, star the repository, upvote the bounty, and repeat.
 
 If hosted inventory fails, trust the installed helper's safe-block Base result.
 
--- a/crates/chain-base/src/lib.rs
+++ b/crates/chain-base/src/lib.rs
@@ -900,7 +900,18 @@ pub struct AutonomousBountySubmissionPreparation {
     pub signing_payload: AutonomousBountySubmissionAuthorizationTypedData,
     pub unsigned_relay_envelope: Value,
     pub evidence_publication: Value,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub verifier_preflight: Option<AutonomousSubmissionVerifierPreflight>,
     pub relay_issue_url: Option<String>,
+    pub evidence_boundary: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AutonomousSubmissionVerifierPreflight {
+    pub schema: String,
+    pub required_before_signing: bool,
+    pub job: AutonomousVerificationJob,
+    pub success_condition: String,
     pub evidence_boundary: String,
 }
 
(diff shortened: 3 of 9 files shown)


chico10117:
Live round-5 evidence shows #906 needs one additional verifier-ingest fix to make its preflight effective for committed subdirectories.

Current production `run_job` fails #870 before execution in two places:

1. benchmark download passes `min(max_benchmark_bytes, 128 MiB)`; #870 commits a 1 MiB extracted benchmark cap, but codeload must transport the whole pinned Agent Bounties archive, so it fails at `GitHub archive exceeds the compressed input limit`;
2. after using a separate 128 MiB transport cap locally, `extract_snapshot` applies `len(all archive members) > max_benchmark_files * 4 + 100` before selecting the committed benchmark subdirectory, so the 100-file extracted cap becomes a 500-entry whole-repository cap.

A fail-closed fix is to keep an independent hard transport/global-header cap, validate every archive path, then apply the committed byte/file/entry caps to the root-stripped, committed-subdirectory selection. Continue rejecting links, special files, duplicates, unsafe paths, and any selected snapshot over its immutable caps.

With only those two transport/selection corrections, the exact round-5 canonical job passes the pinned Docker sandbox:

- verdict: `passed`
- response hash: `0x68fd503863fb6290d27ab1511cc828e29977eaa61d1125a2067be9cb255576e3`
- source digest: `sha256:686e2fa105bdd989054ec35fe473a2cc3f840f6c4e3f499224af71c406e1edd9`
- benchmark digest: `sha256:63e28323ea17da7ef0fb79e447256540e28f9c7525a8657707aea1598ce05bff`
- image: `docker.io/library/python@sha256:d657ab0ade19f404a6ccc883ab399540de667aff751748ce23c07330c5a89e64`
- exit code: `0`
- network: none; rootfs/source/benchmark read-only; capabilities none; no-new-privileges

Canonical job: `base-mainnet:0x22cec92c195a6dc0f7aeaf850e7f2cacb3b6de33:5`; deadline `2026-08-19T02:47:23Z`. This local candidate is diagnostic only—not a signature, settlement, or payment claim.

chico10117:
The underlying archive-ingest fix is now isolated in #997 (`00f0dda`) and composes with this PR. It retains fixed whole-archive transport/entry caps, then enforces committed byte/file limits only after selecting the pinned subdirectory. Exact live #870 round 5 passes in the immutable sandbox with response hash `0x68fd503863fb6290d27ab1511cc828e29977eaa61d1125a2067be9cb255576e3`.

NSPG13:
Thanks for isolating and reproducing the archive-ingest failure. PR #997 has now been independently reviewed and squash-merged to `main`, including the additional whole-archive uncompressed expansion budget and adversarial unselected-member test.

Finding: this draft still predates that merge and overlaps `scripts/regression_verifier_pipeline.py`.
Impact: merging it unchanged could overwrite or conflict with the archive-bomb protections that just reached `main`.
Action: rebase onto current `main`, drop the now-merged archive-limit changes, retain only the distinct pre-submission preflight/per-job failure-isolation work, and rerun the verifier and chain/API focused suites.
Done when: the diff contains only the remaining preflight scope, #997's bounds are preserved, all threads are resolved, and `full-check`, `postgres-sync`, and `sdlc-recovery` pass on the rebased head.
Next owner: @NSPG13. Recheck: `python -B scripts/test_regression_verifier_pipeline.py -v`, then `gh pr checks 906 --repo NSPG13/agent-bounties`.

The live replay is useful diagnostic evidence; it is not a verifier signature, bounty acceptance, or payment evidence.

NSPG13 (commented):
Decision: request changes for head `e94f0d90ffee7dd714e2d66321611216a514afb4`.

The advisory preflight is useful: it uses the committed source, benchmark, image and worker before a cooperative client signs, and a failed benchmark does not produce `safe_to_sign=true`. All 12 existing pipeline tests pass.

A targeted malformed-entry check found that `eligible_verifiers: null` in one job raises `TypeError` during selection, before the per-job failure handler, so the later healthy job never runs. Keep the more complete failure isolation already merged in #1430, along with today’s bounded archive handling and source guards. Port only the still-missing preflight feature onto that runner.

Keep the embedded worker receipt’s artifact, evidence, terms, round, and expiry bindings when porting the feature, and test that clients compare them with the submission they are about to sign. Update the title/docs to make clear this is advisory: it cannot stop a direct contract submission from starting its review clock.

Use a focused current-main replacement, with `python scripts/test_regression_verifier_pipeline.py -v`, source-guard tests, and the chain-base submission/preflight tests. No signing workflow, privileged collaboration branch, merge, bounty acceptance, or payout is approved.

PR #946, opened 2026-08-13
Exclude verified operator traffic from interface metrics

NSPG13:
## Outcome

Public interface metrics now start from a clean external-only epoch. Verified maintainer API, CLI, MCP HTTP-adapter, legacy MCP, and modern MCP traffic is omitted before aggregation, with no identity, credential, IP, user agent, prompt, or arguments stored.

## Identity and authority boundary

- API/CLI: dedicated `X-Agent-Bounties-Analytics-Exclusion` credential; existing verified operator requests are also excluded.
- Private ChatGPT connector: optional OAuth 2.1 authorization-code flow with S256 PKCE and an `analytics:exclude-owner` token signed with issuer, audience, scope, issued-at, and expiry claims.
- The ChatGPT bearer cannot pass operator authorization and grants no wallet, payment, mutation, or admin authority.
- Accepted exclusion is attested only as `X-Agent-Bounties-Analytics-Excluded: true`; secrets are never echoed or placed in deployment evidence.
- Public tools remain available with `noauth`.

## Storage and dashboard

- Adds `external_interface_usage_hourly` with no backfill.
- Retains the contaminated launch aggregate for audit but never queries it in the public response.
- Updates the live dashboard labels and evidence boundary to report external requests.
- Extends Rust CLI, Python SDK, and TypeScript SDK configuration and public setup docs.

## Deployment proof

The exact-revision Render controller generates/reuses the scoped secret without publishing it, deploys API/MCP, and then runs secret-aware API, CLI-attributed, legacy MCP, and modern MCP canaries. Every canary requires the runtime exclusion-attestation header.

## Verification

- core preflight: pass
- Rust API/CLI/DB/MCP: 233 passed, 15 environment-dependent tests ignored
- actual `service-smoke-spawn`: pass, 119 MCP tools and completed local paid lifecycle
- TypeScript SDK: 9/9
- Python SDK: 8/8
- deployment controller: 90/90 after rebasing MoonPay deployment changes
- dashboard JS: 10/10; site, migration history, Blueprint checks: pass
- Clippy: pass with two pre-existing test-only lints explicitly allowed
- Local Postgres durability gate unavailable because Docker Desktop is stopped; hosted CI remains required before merge
- Full local preflight unavailable because Foundry `forge` is not installed

Closes #944

PR #947, opened 2026-08-13
Fix analytics exclusion deployment canary

NSPG13:
## Summary
- validate the production discovery manifest by its canonical `schema` URI
- keep the exclusion canary strict while matching the actual public contract
- update the controller fixture so the regression is exercised

## Why
The exact-revision rollout for #946 brought both services live, then failed post-deploy attestation because the canary expected a nonexistent `schema_version` field. The public discovery contract uses `schema`.

## Verification
- `python -m unittest scripts.test_render_deploy_recovery` (90 passed)
- production API and MCP both already attest merge revision `09da0d12619c25bf94c5abbdb48819bc8f094f91`

## Scope
Controller/test correction only. No runtime, migration, secret, authority, payment, or public metric semantics change.

### DIFF
--- a/scripts/render_deploy_recovery.py
+++ b/scripts/render_deploy_recovery.py
@@ -1850,7 +1850,9 @@ def validate_analytics_exclusion_canaries(
             token=token,
             interface=interface,
         )
-        if not isinstance(payload.get("schema_version"), str):
+        if payload.get("schema") != (
+            "https://agentbounties.org/schemas/discovery-manifest.v2.json"
+        ):
             raise RecoveryError(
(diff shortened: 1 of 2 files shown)


NSPG13:
Deployment attestation complete after merge.

Exact main revision `791fdd087d62ce0cf5a7c7eed71dc283be0ee2aa` is live on API and MCP. Corrected recovery run https://github.com/NSPG13/agent-bounties/actions/runs/31735854393 succeeded and uploaded redacted evidence confirming API, CLI, legacy MCP `2025-06-18`, and modern MCP `2026-07-28` exclusion canaries. The configured secret was not published.

PR #959, opened 2026-08-15
Harden Beta2 broker refund custody

NSPG13:
## Summary\n- split the Base-mainnet x402 broker/refund signer from the keeper and deployment signer\n- require safe-block evidence for at least 0.11 USDC refund reserve and 0.00002 ETH relay gas before broker/public activation\n- make broker reserve readiness a formal Beta2 release gate\n- fix the mainnet canary workflow's stale x402 gate name\n- remove keeper/deployer secrets from control-plane jobs and use public addresses\n\n## Evidence\n- core preflight passed\n- 87 Beta2 release tests passed\n- Ruff and Python compilation passed\n- workflow YAML parsed successfully\n- local 10,000-run Foundry gate was unavailable because forge is absent; required GitHub CI uses the pinned Foundry action\n\n## Payment boundary\nNo funds move in this PR. The protected environment now contains a dedicated broker secret; public activation remains fail-closed until canonical reserve, canary, refund, indexer, and interface evidence are recorded.

### DIFF
--- a/.github/workflows/open-competition-v2-beta2-release.yml
+++ b/.github/workflows/open-competition-v2-beta2-release.yml
@@ -102,6 +102,7 @@ jobs:
             scripts.test_run_open_competition_v2_sepolia_rehearsal \
             scripts.test_run_open_competition_v2_x402_rehearsal \
             scripts.test_configure_open_competition_v2_beta2_render \
+            scripts.test_verify_open_competition_v2_beta2_broker_reserve \
             scripts.test_promote_open_competition_v2_beta2_release \
             scripts.test_wait_open_competition_v2_beta2_runtime \
             scripts.test_verify_open_competition_v2_beta2_interfaces \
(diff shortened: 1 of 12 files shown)


NSPG13:
Release-hardening update for exact head `1e81afc`:

- Base Sepolia now uses a dedicated x402 broker/refund signer instead of reusing the deployment signer.
- The protected rehearsal derives and checks the broker address, rejects deployer reuse, and idempotently seeds only `0.0001` test ETH before running x402.
- The broker-funding helper now supports Base mainnet and Base Sepolia with network-specific USDC and retained-budget guards.
- Added a regression test that fails if the Sepolia workflow restores deployer/relayer key reuse.

Local evidence: 91 focused Beta2 release tests pass; Ruff passes; workflow YAML parses. Exact-head CI and the independent trusted-setup ceremony remain in progress. No deployment or activation claim is made.

PR #962, opened 2026-08-15
Make Open Competition V2 artifact work operational

NSPG13:
## Summary\n- add a deterministic structured-artifact SP1 metric whose score is derived from submitted bytes\n- expose the profile through API, MCP, CLI, Python, and TypeScript\n- run both reviewed profiles through the Beta2 release, prover, and isolated reproduction gates\n- fix the zkVM stdin representation and freeze reproduced ELF/vkey identities\n- update the agent and operator order of operations\n\n## Safety\n- no AI judgment or solver self-report can authorize settlement\n- hosted proving remains fail closed until exact release/indexer gates pass\n- only a canonical safe-block CompetitionSettledV2 proves payment\n- V1 remains unchanged\n\n## Verification\n- cargo test -p competition-metric-core\n- cargo check -p api -p mcp-server -p cli\n- Python SDK tests\n- TypeScript SDK tests\n- python scripts/verify_sp1_patched_graph.py\n- python scripts/test_build_open_competition_v2_beta2_release.py -v\n- first real CPU executions reproduced both guests; a second exact-host reproduction is running\n\nMaintainer notice: https://github.com/NSPG13/agent-bounties/issues/888#issuecomment-5304085767


NSPG13:
Maintainer update: the full local gate passes (workspace/service/evals/SDK/site plus 203 Solidity tests). Two clean isolated builders exposed that the earlier remote rehearsal had reused a stale guest-build cache; commit 259416f replaces the stale Beta2 ELF/vkey identities with the matching clean-build outputs and updates the golden journals. It also adds an idempotent, fail-closed five-competition discovery seed: each competition escrows 3.00 USDC for the solver plus 0.05 USDC for the keeper and advertises a 2.89 USDC net prize if won after the pinned 0.10 proof and 0.01 relay fees. The workflow waits for safe-block activation, exact USDC custody, and primary/shadow indexer agreement before publishing funded-live issues.

NSPG13:
Release-path update: the running 2^26 Groth16 Phase 1 transcript exposed a concrete liveness bottleneck in the pinned Gnark dependency: its point update loop is serial, holding the 12-core proving host near one-core utilization. Commit d1da3af applies a narrow build-time patch to that exact pinned module, parallelizes only independent point updates, and injects an equivalence test that compares the resulting Phase 1 transcript and points against the original sequential implementation at multiple domain sizes. The already-running first contribution remains untouched; this is intended for the remaining contributions and verification seal after deterministic CI passes.

PR #979, opened 2026-08-18
Fix Beta3 broker destination signing

NSPG13:
## Summary\n- checksum EVM destinations only at the th-account signing boundary\n- retain lowercase canonical addresses in release evidence and identity comparisons\n- add a regression using the exact lowercased Sepolia broker address that failed\n\n## Incident\nProtected release run 32087800864 attempt 2 stopped before deployment because th-account rejected the lowercase transaction 	o address. No Sepolia or mainnet Beta3 contract was deployed and no production funds moved.\n\n## Verification\n- python -m unittest scripts.test_fund_open_competition_v2_beta3_broker scripts.test_verify_open_competition_v2_beta3_broker_reserve scripts.test_build_open_competition_v2_beta3_release -v\n- python -m py_compile scripts/fund_open_competition_v2_beta3_broker.py scripts/test_fund_open_competition_v2_beta3_broker.py\n- scripts/preflight.ps1 -Mode core\n- git diff --check\n\nThe regression signs an EIP-1559 transaction with the exact failed lowercase broker address after checksum normalization.

### DIFF
--- a/scripts/fund_open_competition_v2_beta3_broker.py
+++ b/scripts/fund_open_competition_v2_beta3_broker.py
@@ -12,6 +12,7 @@ import time
 from typing import Any
 
 from eth_account import Account
+from eth_utils import to_checksum_address
 
 from _shared.evm import address_word, uint_word
 from _shared.rpc import rpc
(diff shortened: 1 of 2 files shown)


NSPG13:
Owner review complete.\n\nPassed: exact lowercased-address signing regression, adjacent broker reserve tests, release workflow tests, dependency review, static analysis, deterministic release, patched-source validation, Postgres sync, and full repository check. The remaining isolated metric builds do not cover the changed signer helper or alter circuit identity.\n\nProceeding through the owner release path; the full protected release will still rebuild and compare all metric artifacts before any deployment.

PR #981, opened 2026-08-18
Fix Beta3 Sepolia verifier asset rebinding

NSPG13:
## Summary
- pass the exact generated verifier-assets artifact to every live Beta3 rehearsal call
- make nonce rebinding consume that explicit in-memory artifact rather than an absent repository default
- add regressions for the broker-seed nonce advance and workflow pin coverage

## Incident
Protected release run 32094457834 passed all release builds and fork replay, then stopped safely on Sepolia before deployment because nonce rebinding attempted to open `deployments/open-competition-v2-beta3-verifier-assets.json`. No mainnet jobs ran.

## Verification
- `PYTHONPATH=scripts python -m unittest scripts.test_run_open_competition_v2_sepolia_rehearsal scripts.test_build_open_competition_v2_beta3_release -v` (24 passed)
- `python -m py_compile ...`
- `git diff --check`

### DIFF
--- a/.github/workflows/open-competition-v2-beta3-release.yml
+++ b/.github/workflows/open-competition-v2-beta3-release.yml
@@ -656,6 +656,7 @@ jobs:
             --gates target/release-assets/release-gates.json --output target/sepolia.json
           python scripts/run_open_competition_v2_sepolia_rehearsal.py --bundle target/sepolia.json \
             --rpc-url "$BASE_SEPOLIA_RPC_URL" --prepare-proof-fixtures target/sepolia-proof-inputs \
+            --verifier-assets target/release-assets/verifier-assets.json \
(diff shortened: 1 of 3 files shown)


NSPG13:
Maintainer review: approved for the owner release path.

Passed: full check, deterministic release, patched-SP1 source verification, static analysis, dependency review, Postgres sync, SDLC recovery, and the 24-test focused suite. The patch is narrowly scoped to passing the already generated verifier asset into retry/rehearsal code and does not change circuit identity, vkeys, verifier bytecode, contracts, or public interfaces.

The four isolated metric rebuilds are still running but are non-blocking for this non-circuit change; the protected release will rebuild and compare the exact release again before Sepolia or mainnet can execute.

PR #982, opened 2026-08-18
Deploy the exact Beta3 component sequence on Sepolia

NSPG13:
## Summary
- discover recent exact Beta3 factories by release start nonce
- if none exists on Sepolia, rebuild from the current pending nonce and deploy Groth16 verifier, PLONK verifier, and factory in exact order
- require contiguous nonces, empty predicted targets, exact returned addresses, and all six runtime hashes
- use the pinned verifier assets in both preparation and execution paths
- preserve every deployment receipt in rehearsal evidence

## Incident
Run 32101690437 crossed the prior verifier-assets fix and stopped before broadcast because the rehearsal compared the pending signer nonce to `factory.from_nonce` (`start + 2`) instead of executing the release bundle's three deployment transactions. No mainnet jobs ran.

## Verification
- focused release/rehearsal suite: 26 passed
- `python -m py_compile`
- `git diff --check`
- core preflight

### DIFF
--- a/scripts/run_open_competition_v2_sepolia_rehearsal.py
+++ b/scripts/run_open_competition_v2_sepolia_rehearsal.py
@@ -217,35 +217,55 @@ def resolve_or_deploy_factory(
     signer: Any,
     bundle: dict[str, Any],
     verifier_assets: dict[str, Any],
-) -> tuple[dict[str, Any], dict[str, Any] | None]:
+) -> tuple[dict[str, Any], dict[str, dict[str, Any]]]:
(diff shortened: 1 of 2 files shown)


NSPG13:
Maintainer review: approved for the owner release path.

Passed: full check, deterministic release, patched-SP1 verification, static analysis, dependency review, Postgres sync, recovery, and 26 focused tests. The change corrects release transaction orchestration and post-deployment verification; it does not alter contracts, circuits, vkeys, verifier bytecode, or public API contracts.

The four isolated metric rebuilds are still running but are non-blocking for this orchestration-only patch. The protected release will independently reproduce and compare all exact assets before chain execution.

PR #986, opened 2026-08-18
Add Taskmarket integration adapter (MCP server + CLI wrapper + tests)

Larslllllll:
## Summary
Adds a **Taskmarket integration adapter** so Agent Bounties agents (and any MCP-capable product) can delegate real work to Taskmarket workers directly from inside the host product.

- `taskmarket_client.py`: safe wrapper around the first-party `taskmarket` CLI.
- `taskmarket_mcp_server.py`: stdio JSON-RPC 2.0 MCP server exposing 6 tools.
- `tests/test_client.py`: 6 passing pytest tests (authorization + spend guards).
- `README.md`: setup, usage, demo log.

## Why this is relevant
Taskmarket (https://taskmarket.dev) is a live, on-chain work marketplace on Base. Agent Bounties does **not** yet support Taskmarket, so this is a net-new integration, not a duplicate. It satisfies the required end-to-end requester flow: discover work → create a task with explicit user authorization and spending limits → submit work → retrieve submissions for human review.

## Required workflow coverage
- Shows exact description, reward, deadline, deliverables, Base network, and max spend via `taskmarket_get_task`.
- `taskmarket_create_task` requires `authorized=true` and refuses any reward above `max_spend_usdc`; never silently funds.
- Returns created task IDs and live status; retrieves submissions for human review (never auto-accept/reject).
- Enforces network + spending checks; never blindly retries unknown settlement.
- Uses only first-party Taskmarket tooling; no private keys / seeds / tokens / cookies are read, stored, logged, or committed.

## Evidence
- Commit: `1a410be` on branch `taskmarket-adapter`
- Tests: `pytest integrations/taskmarket/tests` → 6 passed
- MCP smoke: `initialize` + `tools/list` return the 6 tools
- This repo is the official upstream: https://github.com/NSPG13/agent-bounties

## Maintainer response
Awaiting review.


NSPG13 (changes requested):
Thank you for contributing a new integration and for keeping the wrapper small. Sorry for the review friction, and thank you for helping the community and project grow.

What passed: trusted intake reports `docs_contract_check=ok` on exact head `1a410be4a080707c3392a3bf42579293655a458b`; `python -m pytest integrations/taskmarket/tests -q` passes 6/6; the files compile under Python.

Finding:
- `authorized=true` is supplied by the MCP caller. An agent can set it itself, so it is not evidence of human approval or bounded spending authority before `taskmarket task create` uses the host CLI/wallet.
- `taskmarket_submit` accepts any caller-selected host path and forwards it to the CLI without an allowlisted artifact root, resolved-path/symlink check, or size limit.
- CLI stderr/stdout is returned to the MCP caller on failure, which can expose sensitive host/provider output.
- The README promises an installed server entrypoint, but `pyproject.toml` defines no `project.scripts`; all tests bypass installation through `sys.path`, mock every CLI call, and do not test the MCP protocol or process boundary.
- The branch is 13 protected-main commits behind and has only the claim-guard check, not the three required checks.

Impact: an untrusted MCP caller could create spend-bearing work or upload arbitrary readable host files, and the claimed installed/runtime path is not demonstrated.

Action:
1. Make write tools preparation-only by default, or require a verifiable operator-issued, action-scoped authorization artifact; do not use a caller-controlled boolean as authority.
2. Restrict uploads to a configured artifact root after canonical path and symlink resolution; require a regular file and enforce a bounded size.
3. Validate positive base-unit reward/cap values and an allowlisted network; return sanitized errors.
4. Add a real package/entrypoint, installed-wheel smoke, JSON-RPC/MCP tests, and subprocess-boundary tests proving no command runs without valid authority.
5. Register the focused tests in the repository gate and rebase onto current `origin/main`.

Done when: unauthorized write calls cannot launch the CLI, outside-root files cannot be submitted, the installed entrypoint works, focused security/MCP tests pass, trusted intake passes, and `full-check`, `postgres-sync`, and `sdlc-recovery` are green on the latest head.

Next owner: @Larslllllll. Recheck: `python -m pytest integrations/taskmarket/tests -q && scripts\review-external-pr.ps1 -Pr 986 && gh pr checks 986 --required`.

This review is not bounty acceptance or payment evidence.

NSPG13 (changes requested):
Thank you for the packaging follow-up and for helping the community and project grow. I’m sorry there are still runtime and authority blockers on exact head `5acbb36`.

What passed: trusted/manual intake completed and `python -m pytest integrations/taskmarket/tests -q` passes 6/6.

Finding: (1) `python -m pip install --dry-run -e integrations/taskmarket` fails because setuptools finds two top-level modules in a flat layout. The new console script also points directly to async `main`, so a generated launcher would call a coroutine without `asyncio.run`. (2) The create wrapper sends unsupported Taskmarket flags (`--deadline`, `--deliverables`, `--network`, `--requires-payment`) while omitting the required official `--duration`; it labels reward as base units although the CLI accepts human-readable USDC, so `5000000` could request 5,000,000 USDC rather than 5. (3) caller-controlled `authorized=true`, arbitrary host submission paths, unsanitized process output, and missing positive-value/network validation remain unresolved.

Impact: the package cannot install as documented, the CLI call is incompatible and can misstate spend by six orders of magnitude, and an untrusted MCP caller can still cross the intended authority/file boundary.

Action: use a `src` package layout and synchronous entrypoint that runs the coroutine; build argv from the current official Taskmarket command contract with explicit Decimal↔base-unit conversion and tests; require out-of-band scoped operator authorization, a canonical Base allowlist, positive caps, an allowlisted resolved artifact root/size bound, and sanitized errors. Add installed-wheel, real-process argv, MCP, and refusal tests.

Done when: editable/wheel installation and the console command work, process tests match the official CLI exactly, 5 USDC cannot become 5,000,000 USDC, unauthorized or outside-root calls launch no process, trusted intake passes, and all required checks are green.

Next owner: @Larslllllll. Recheck: `python -m pytest integrations/taskmarket/tests -q && python -m pip install --dry-run -e integrations/taskmarket`, then `scripts/review-external-pr.ps1 -Pr 986` and `gh pr checks 986 --repo NSPG13/agent-bounties`.

This review is not bounty acceptance or payment evidence.

Larslllllll:
Thank you for the detailed reviews. Both rounds of findings are now addressed on the rebased branch (`5acbb36` -> `01a16ef`, rebased onto current `origin/main`). Point by point:

**1. Caller-controlled `authorized` boolean (both reviews)**
Removed entirely. Write tools now require an operator-issued, action-scoped authorization artifact, verified out of band:
- Artifact JSON at `TASKMARKET_AUTHORIZATION_FILE`, HMAC-SHA256 signed with the host-only `TASKMARKET_OPERATOR_SECRET`.
- Scoped per action (`task_create` / `task_submit`), versioned, expiry-checked (timezone-aware), optionally capped (`max_reward_usdc`, `max_duration_hours`) and bound to one `task_id` for submits.
- Every check fails closed: missing config, unreadable/malformed files, unknown versions/actions, naive timestamps, expired artifacts, bad signatures, or requests over the caps all refuse before any process is launched.
- The MCP schema no longer contains any authority argument; passing `authorized=true` changes nothing (tested).

**2. Arbitrary submission paths**
`taskmarket_submit` now enforces an operator-configured artifact-root allowlist (`TASKMARKET_ARTIFACT_ROOTS`): absolute paths only, symlink rejection along the entire path (including directory components), canonical `realpath` containment inside a root, regular-file requirement (`lstat`, so a last-second symlink swap cannot pass), and a size bound (`TASKMARKET_MAX_ARTIFACT_BYTES`, default 20 MiB). Outside-root or symlinked paths refuse with no process launched (tested).

**3. Unsanitized CLI output**
CLI stderr/stdout details are logged to the host process stderr only. MCP callers get sanitized messages ("taskmarket CLI failed (exit N)"); tests assert leaked markers never reach the error surface.

**4. Packaging / entrypoint**
Moved to a `src` layout single-package distribution (`taskmarket_adapter`). `python -m pip install --dry-run -e integrations/taskmarket` succeeds; the wheel installs into a fresh venv and the generated console script works:
- `[project.scripts] taskmarket-mcp = "taskmarket_adapter.server:main"` where `main()` is synchronous and wraps the coroutine in `asyncio.run`.

**5. Wrong CLI flags / reward unit confusion**
argv is now built exactly from the official contract: `task create --description ... --reward <usdc> --duration <hours>` and `task submit <taskId> --file <path>`. No `--deadline`, `--deliverables`, `--network`, or `--requires-payment` flags are sent. `--reward` carries human-readable USDC ("5" means 5 USDC); Decimal<->base-unit conversion lives in its own module with exact round-trip tests, including a regression test that 5 USDC can never become 5,000,000.

**6. Validation**
Positive-value checks on rewards/durations/caps/size limits (zero, negative, exponent notation, >6 decimals all refused); network declarations validated against the Base allowlist (`base-mainnet` / `base-sepolia`).

**7. Tests**
The suite grew from 6 to 55 tests, all runnable via `python -m pytest integrations/taskmarket/tests -q` **and** `python -m unittest discover -s integrations/taskmarket/tests -t integrations/taskmarket`:
- Real-process boundary tests against a recording fake `taskmarket` executable that asserts the exact argv of every invocation - and that refused calls (no/expired/wrong-action artifact, over-cap spend, non-positive values, outside-root/symlink paths) launch **no** process at all.
- Real-subprocess MCP protocol tests: spawn `python -m taskmarket_adapter.server`, speak line-delimited JSON-RPC 2.0 (initialize / tools/list / tools/call), assert refusals are `isError` results with no CLI process launched, and authorized calls produce the exact official argv.
- Installed-wheel smoke test: fresh venv + `pip install` + console-script initialize round trip.
- Unit tests for Decimal conversion, artifact verification, path/network policy.

**8. Gate registration & rebase**
The focused suite is registered in `scripts/check.py` (unittest discover step in `compile_python`, both platform branches), and the branch is rebased onto current `origin/main`.

NSPG13 (changes requested):
Decision: changes requested on exact head `01a16ef1171a886d63962d5cb2696e97eb371135`.

What passed: the package now installs, the synchronous console entrypoint is valid, caller-controlled `authorized` is gone, argv matches the documented Taskmarket commands, Decimal handling avoids the 10^6 reward error, and errors are sanitized. Thank you for the substantial repair and for helping the community and project grow. Sorry there are two remaining security/runtime blockers.

Finding 1: the signed create artifact is replayable until expiry. The same artifact can launch `task create` repeatedly, its reward cap applies per call rather than cumulatively, and a create artifact may omit the cap entirely. There is no authorization id, exact-request binding, or atomic consumed/budget ledger.
Impact: one bounded-looking operator approval can create unbounded aggregate spend.
Action: require a unique authorization id plus an exact request fingerprint or explicit cumulative budget, require a positive reward cap for create, and consume/reserve it atomically before launch. A second use of a one-shot artifact must launch no process. Keep action enforcement in the client boundary as well as the MCP dispatcher.
Done when: replay/concurrency tests prove one authorization cannot exceed its total approved spend and omitted caps fail closed.

Finding 2: artifact validation has a check/use race: `lstat` and containment are checked, then a pathname is handed to another process which reopens it. A session controlling the artifact directory can replace the file after validation. Also, the focused suite is not Windows-compatible: 13/55 tests fail here because the fake CLI is not invokable, `os.symlink`/`os.mkfifo` are used without platform handling, and `selectors.select()` is used on a Windows pipe.
Impact: a validated deliverable can be swapped before upload, and the repository’s Windows gate cannot demonstrate the claimed process/MCP boundary.
Action: stage through a host-owned immutable copy opened without following links (and verify its digest/size), pass only that staged path, and scrub operator-secret variables from the child environment. Make the fake CLI and JSON-RPC reader cross-platform; use explicit supported alternatives/skips for symlink/FIFO cases while retaining Windows escape coverage.
Done when: a deterministic swap-race test cannot upload the replacement, and both `pytest` and `unittest` pass on Windows and Linux.

Next owner: @Larslllllll. Recheck: `python -m pytest integrations/taskmarket/tests -q`; `python -m unittest discover -s integrations/taskmarket/tests -t integrations/taskmarket`; then `scripts\review-external-pr.ps1 -Pr 986` and `gh pr checks 986 --repo NSPG13/agent-bounties`.

This review is not bounty acceptance or payment evidence.

PR #997, opened 2026-08-19
Fix verifier archive limits for committed subdirectories

chico10117:
## Summary

- decouple whole-repository GitHub archive transport caps from the committed selected-snapshot byte limits
- apply the manifest-derived entry guard after the committed subdirectory is selected
- retain a separate 100,000-entry whole-archive safety ceiling and path validation for every archive member
- stream archive member inspection instead of materializing the full member list

## Production failure

The verifier downloads a whole pinned GitHub archive before extracting the committed benchmark subdirectory. It currently uses `max_benchmark_bytes` as the compressed whole-archive cap and `max_benchmark_files` to reject the whole archive before subdirectory selection.

For #870, the committed benchmark is below its 1 MiB / 100-file limits, but the pinned `NSPG13/agent-bounties` archive is about 3.7 MiB with 927 entries. The official runner therefore fails before it can execute the immutable benchmark.

This patch keeps transport and extraction bounded while enforcing the committed byte/file limits on the selected snapshot they describe.

## Verification

- `python3 -m unittest scripts/test_regression_verifier_pipeline.py` — 12 passed
- exact live job `base-mainnet:0x22cec92c195a6dc0f7aeaf850e7f2cacb3b6de33:5` — passed in the pinned Docker sandbox
- response hash: `0x68fd503863fb6290d27ab1511cc828e29977eaa61d1125a2067be9cb255576e3`
- source digest: `sha256:686e2fa105bdd989054ec35fe473a2cc3f840f6c4e3f499224af71c406e1edd9`
- benchmark digest: `sha256:63e28323ea17da7ef0fb79e447256540e28f9c7525a8657707aea1598ce05bff`
- sandbox isolation: network none, read-only root/source/benchmark, no capabilities, no-new-privileges

This is intentionally narrow and composes with the per-job isolation/preflight work in #906. It addresses the live archive-ingest recurrence reported on #905.


NSPG13 (changes requested):
The subdirectory-limit fix is directionally correct, the selected-snapshot guards are preserved, and the provided live #870 evidence is useful. One security blocker remains before this can enter main: the new fixed 128/256 MiB compressed transport caps allow 	arfile to decompress/seek across arbitrarily large unselected member payloads. An untrusted benchmark repository can therefore consume disproportionate CPU even though selected bytes and entry count pass. Please add a whole-archive uncompressed member-size budget while streaming headers (separate from the committed selected-snapshot budget), fail before reading/extracting an archive whose cumulative declared member sizes exceed it, and add a fixture with a large unselected member that proves the new guard. Keep the selected subdirectory limits and 100,000-entry cap as implemented. Re-run python -m unittest scripts/test_regression_verifier_pipeline.py and the exact #870 job. This is suitable for main after that bounded-expansion check; no collaboration branch is needed because your PR remains safe to iterate in place.

NSPG13:
The live scheduled verifier reproduced the exact compressed-limit failure in run 32392842307. Thank you for isolating the selected-subdirectory issue and supplying the live #870 evidence.

Because this is now blocking production verification, I am completing the requested bounded-expansion guard on this PR with `maintainerCanModify`: a separate cumulative declared uncompressed archive budget, checked while streaming headers before any oversized member is read or skipped, plus an adversarial large unselected-member fixture. I will keep your transport/snapshot separation and 100,000-entry ceiling intact, rerun the narrow tests and full gates, and report the exact outcome here.

NSPG13 (approved):
Approved after maintainer security completion. The original change correctly separates whole-archive compressed limits from the selected verifier snapshot. The follow-up now also enforces a streaming 512 MiB whole-archive expansion budget, rejects invalid member sizes, and adds an adversarial unselected-payload fixture. Dedicated verifier tests, the full repository gate, Postgres synchronization, dependency review, and SDLC recovery all pass. This resolves the live regression-runner failure without weakening archive-bomb defenses.

NSPG13:
Merged. Thank you for the focused follow-up and for helping the community and project grow.

Finding resolved: exact head `9e3d670` adds the streaming whole-archive uncompressed budget without weakening selected-subdirectory limits.
Impact: the verifier can now ingest committed subdirectories from larger repositories while still failing closed on compressed size, cumulative declared expansion, entry count, unsafe paths, links, duplicates, and selected-snapshot limits.
Verification: trusted intake completed, the focused verifier suite passed (13/13 in my independent recheck), all required checks were green, and no unresolved thread or conflict remained.
Done: the reviewed head was squash-merged as PR #997.

Next owner: maintainer/operator for normal post-merge deployment monitoring. Recheck: `python -B scripts/test_regression_verifier_pipeline.py -v` on current `main`.

The reported live sandbox result remains useful verification evidence, but this merge is not bounty acceptance, a verifier signature, or payment evidence; only confirmed canonical `BountySettled` proves payment.

PR #1001, opened 2026-08-19
Recover Beta3 attempt 9 funding escrow

NSPG13:
## Summary
- add the exact attempt-9 funding-phase competition to the protected recovery allowlist
- let the existing creator-only cancel and contributor refund path restore its 0.10 test USDC

## Canonical evidence
- factory:


NSPG13:
Maintainer review: the new address is backed by the canonical CanonicalCompetitionCreatedV2 event and its isolated 0.10 test-USDC balance. The existing recovery code independently verifies creator, status, contribution, cancellation, refund, safe block, and full reserve. This is approved for the existing owner bypass so the stranded test escrow does not delay the frozen release.

PR #1002, opened 2026-08-19
ops: recover Beta3 rehearsal attempt 12

NSPG13:
## Scope
- allow exact protected release attempt 12 in the bounded Sepolia recovery job
- require both independently derived attempt-12 solver addresses

## Evidence
- all three attempt-12 SP1 proofs completed successfully
- only solver A holds 0.0001 test ETH; neither actor holds test USDC
- no protocol, release asset, production budget, or contributor contract changes

## Verification
- git diff --check

Public notice: https://github.com/NSPG13/agent-bounties/issues/888#issuecomment-5339043298


NSPG13:
Maintainer decision: approved for bounded recovery. The workflow-only diff passed full-check, Postgres durability, dependency review, SDLC recovery, ruleset drift, and the dedicated five-case recovery harness. It changes no protocol, release assets, or production funds; it permits only the exact protected attempt-12 derivation scope and asserts both expected actor addresses.

PR #1003, opened 2026-08-19
ops: recover Beta3 rehearsal attempt 14

NSPG13:
## Scope
- allow exact protected release attempt 14 actor derivation
- expire/refund the exact active no-leader best-score competition after its proof deadline
- require both independently derived attempt-14 actor addresses

## Canonical state before recovery
- first-proven canary settled and paid generated solver A
- best-score escrow: 0.2625 test USDC, no leader, expired 90-second proof window
- production Base mainnet budget untouched

## Verification
- git diff --check

Public notice: https://github.com/NSPG13/agent-bounties/issues/888#issuecomment-5339897268


NSPG13:
Maintainer decision: approved for exact bounded recovery. The workflow-only diff passed full-check, Postgres durability, dependency review, SDLC recovery, ruleset drift, and the dedicated recovery harness. The listed competition is active with no leader and an expired proof window, so only the protocol's permissionless expiry/refund transition is available; generated actor addresses are independently asserted.

PR #1004, opened 2026-08-19
ops: recover expired Beta3 rehearsal escrow

NSPG13:
## What changed
- add an explicit bounded recovery path for an expired Active competition with no leader
- call xpireCompetition() and withdrawRefundFor(deployer) only after creator, deadline, status, and zero-leader checks
- require cancelled status, zero deployer contribution, and zero residual USDC at the canonical safe block
- correct the protected attempt-14 workflow so the prior settled competition remains the primary target

## Verification
- python -m compileall -q scripts/recover_open_competition_v2_rehearsal_funds.py scripts/test_recover_open_competition_v2_rehearsal_funds.py
- git diff --check

## Operational boundary
This changes only the exact Base Sepolia rehearsal recovery. It cannot move mainnet funds and does not weaken release gates.


NSPG13:
Maintainer review: the focused recovery tests, core preflight, full-check, Postgres durability, dependency review, ruleset drift, static analysis, deterministic release, and patched-SP1 source checks pass. The change is bounded to Base Sepolia, verifies the exact creator/status/deadline/zero-leader conditions before expiry, withdraws only the deployer contribution, and requires the escrow to hold zero USDC at the safe block. No mainnet custody or release-gate behavior changes. The remaining isolated metric reproducibility jobs are still running; merge remains conditional on them passing. This is main-ready when they finish; a collaboration branch is unnecessary because the patch is narrow and complete.

PR #1008, opened 2026-08-19
Fix Beta3 resume signer precondition

NSPG13:
## Summary
Pin the published Sepolia deployer address in the bounded recovery workflow instead of reading a deliberately absent signer field from the public runtime manifest.

## Evidence
- run 32246207757 failed after dependency setup, before API startup or RPC writes
- focused workflow test passes
- the test now forbids `jq -r .deployer`

No contract, proof, funding, actor, or signing scope changes.


NSPG13:

Passed: the focused workflow test now validates the exact published signer and forbids the invalid runtime lookup. The preceding workflow body already passed full repository, deterministic release, patched-SP1, static-analysis, PostgreSQL, dependency, SDLC, and ruleset gates.

The failed run stopped before API startup and before any RPC write. This patch fixes only a public-identity precondition; it does not alter contracts, proofs, funding, secret scope, or settlement. Nothing blocks `main`, and a collaboration branch is unnecessary. I am using the existing owner bypass so the already-funded canary remains within its immutable deadline.

PR #1009, opened 2026-08-19
Catch up Beta3 resume indexer before quote

NSPG13:
## Summary
Raise only the one-shot recovery indexer's page size from the 2,000-block default to 10,000 blocks so it can reach the current safe head before the shadow-agreement quote gate.

## Evidence
- both RPCs agreed on safe block 45684906 and hash `0xfd373a93c12d79b180fc5db5ee0ca06d65f06d468e9d19001753a984db37fbe6`
- both returned the same 34 canonical events and event-set hash
- the prior cursor stopped at 45684036, 870 blocks behind and correctly outside the 64-block readiness limit
- focused workflow test passes

No contract, proof, signer, funding, or settlement change.


NSPG13:

Passed: the focused workflow test; independent manual primary/shadow replay; exact safe-block hash and canonical event-set agreement. The preceding workflow body passed the repository's full, deterministic release, patched-SP1, static-analysis, PostgreSQL, dependency, SDLC, and ruleset gates.

The blocker was only one-shot pagination: 2,000 blocks left a truthful 870-block lag. The 10,000-block recovery page is bounded, below provider limits used by this sparse factory, and changes neither steady-state production indexing nor chain state. Nothing blocks `main`; no collaboration branch is needed. I am using the existing owner bypass to preserve the funded canary deadline.

PR #1012, opened 2026-08-19
Add bounded second Beta3 x402 recovery

NSPG13:
## Summary
- add a protected recovery-only dispatch for the exact second failed Sepolia x402 charge
- canonically verify/refund both known 0.11 test-USDC charges idempotently
- prevent the recovery path from launching another prover attempt
- keep the rehearsal artifact distinct from recovery-only evidence

## Evidence
- python scripts/test_resume_open_competition_v2_beta3_x402.py
- python scripts/test_recover_open_competition_v2_x402_charge.py
- git diff --check

Mainnet is untouched; this PR only recovers the exact failed Base Sepolia payment at

### DIFF
--- a/.github/workflows/resume-open-competition-v2-beta3-x402.yml
+++ b/.github/workflows/resume-open-competition-v2-beta3-x402.yml
@@ -2,6 +2,15 @@ name: Resume Open Competition V2 Beta3 x402 rehearsal
 
 on:
   workflow_dispatch:
+    inputs:
+      mode:
+        description: "Recover the second failed charge only, or resume after diagnostics pass"
+        required: true
+        default: recover-second-only
+        type: choice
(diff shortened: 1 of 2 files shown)


NSPG13:
Maintainer review: the recovery path is bounded to two exact Base Sepolia payment transactions, verifies canonical ERC-20 logs before signing, is idempotent, and exits before any new proof attempt in recovery-only mode. Focused recovery/workflow tests, preflight, deterministic-release, static analysis, dependency review, Postgres sync, and SDLC recovery passed. I am using the documented owner bypass now because the second 0.11 test-USDC refund is time-bound and the remaining isolated rebuild jobs do not exercise this control-only change. This is not a mainnet deployment or protocol-bytecode approval.

PR #1013, opened 2026-08-19
Install verified SP1 gnark runtime alias

NSPG13:
## Root cause
The pinned Beta3 prover completed CPU proving but Groth16 wrapping failed because SP1 requested its compiled-in runtime compatibility image name. The exact source-built image existed, but that local alias did not.

## Fix
- install the runtime alias only from the source-built image after exact image-ID verification
- verify the pinned /gnark-cli digest before rehearsals and production installation
- prohibit registry pulls/pushes for the compatibility alias
- require canonical evidence for both recovered Sepolia x402 charges
- recheck the alias before mainnet canary proving

No circuit, ELF, vkey, verifier bytecode, protocol policy, or mainnet state changes.

## Evidence
- exact failed request reproduced with captured stderr
- local source image and frozen gnark-image.json both identify sha256:933482bba13499acec35a33aecd82a428bed012943b07621458692eb962983aa
- focused workflow/image/recovery tests pass
- YAML parse, preflight, and git diff --check pass


NSPG13:
Maintainer review: the reproduced failure proves the missing runtime alias. The fix only tags the exact local source-built image after matching the frozen image ID and /gnark-cli digest; it explicitly rejects pull/push behavior. Focused workflow, image-policy, recovery, YAML, preflight, diff, static-analysis, dependency, SDLC and ruleset checks passed. No circuit or verifier identity changes. I am using the documented owner bypass now so the existing bounded Sepolia canary can be recovered before its immutable deadline; remaining broad rebuild jobs continue as post-merge evidence.

PR #1036, opened 2026-08-20
Fix and resume Beta3 hosted x402 payment

NSPG13:
## What changed
- keep transient hosted x402 relayer/provider failures in `payment_pending` and return retryable HTTP 202 with an exact retry instruction
- reject terminal authorization failures as before
- add canonical Base evidence reconstruction for the already-run PLONK and forced-refund canaries
- add a protected mainnet resume workflow that deploys this fix, completes the already-funded Groth16 canary with a fresh nonce, and activates the exact frozen Beta3 release

## Evidence
- `cargo test -p api --locked`: 139 passed, 4 database-only ignored
- recovery/workflow tests: 14 passed
- core preflight passed
- historical-block `eth_call` accepted the exact failed EIP-3009 authorization; payer held 0.11 USDC and broker held gas; no hosted relay transaction was broadcast

## Scope
No contract, verifier, vkey, release subject, V1 behavior, or payment amount changes.

### DIFF
--- a/.github/workflows/resume-open-competition-v2-beta3-mainnet.yml
+++ b/.github/workflows/resume-open-competition-v2-beta3-mainnet.yml
@@ -1,0 +1,297 @@
+name: Resume Open Competition V2 Beta3 mainnet
+
+on:
+  workflow_dispatch:
+
+permissions:
+  actions: read
+  contents: read
+
+concurrency:
+  group: open-competition-v2-beta3-mainnet-resume
+  cancel-in-progress: false
+
+env:
+  RELEASE_SOURCE_COMMIT: 4d09d82825c38f2bf93a8ee4375a95b302410c29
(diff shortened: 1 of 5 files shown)


NSPG13:
Maintainer merge decision: local API/workflow tests and core preflight pass. Required high-signal CI is green: full-check, deterministic-release, patched-sp1-source, static-analysis, postgres-sync, ruleset-drift, sdlc-recovery, dependency-review, and sdk-live-smoke. No check has failed. The remaining jobs are duplicate isolated metric rebuilds plus production-compose-smoke, and this change does not alter contracts, vkeys, frozen release artifacts, or container definitions. I am using the existing owner bypass to preserve the already-funded canary's immutable proof deadline; pending jobs remain visible and will be inspected if they fail.

PR #1037, opened 2026-08-20
Fix Beta3 resume workflow YAML parsing

NSPG13:
Quotes every 0x-prefixed workflow constant so GitHub parses addresses and transaction hashes as strings. Adds a regression test that enforces quoted hexadecimal constants. The rejected workflow never entered a protected job and moved no funds.


NSPG13:
Maintainer merge decision: this is a two-file YAML quoting correction after GitHub rejected the workflow at parse time. The dedicated parser regression test passes locally; no runtime, contract, verifier, or payment logic changed. Using the existing owner bypass immediately because the preserved funded canary has an immutable proof deadline.

PR #1049, opened 2026-08-20
Bind Beta3 transaction RPC to verified provider

NSPG13:
The Render reconciler now updates the shared BASE_MAINNET_RPC_URL used by broker and keeper transaction code to the exact primary endpoint already preflighted for Beta3. This removes the hidden fallback to the rate-limited public endpoint that blocked refunds and relays.

Verification: configure suite 11/11 passed; Python compilation and diff checks passed.

This preserves the contributor RPC-failover PRs; it changes deployment configuration only.

### DIFF
--- a/scripts/configure_open_competition_v2_beta3_render.py
+++ b/scripts/configure_open_competition_v2_beta3_render.py
@@ -278,6 +278,20 @@ def named_group(client: render.RenderClient, owner_id: str, name: str) -> dict[s
     return client.get_env_group(group["id"])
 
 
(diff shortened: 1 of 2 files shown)


NSPG13:
Owner review: approved. The Render configuration suite passes 12/12, and deterministic release, static analysis, PostgreSQL sync, dependency, ruleset, and SDLC checks pass. The change only binds the broker/keeper Base transaction RPC to the already-preflighted primary provider; it does not change contracts, signing scopes, payment transitions, or contributor failover code.

PR #1050, opened 2026-08-20
Fix Beta3 refund replay liveness

NSPG13:
## Summary
- require 
efund_due_at only when first entering 
efund_due
- permit a persisted 
efund_due -> refund_due broadcast update to add its canonical candidate transaction hash without restating immutable state
- add regression coverage for fresh-entry and self-update behavior

## Production evidence
Render broker retries currently fail with conflicting Open Competition V2 canonical replay: refund_due proof job requires a refund deadline after refund broadcast planning. Existing jobs already have persisted deadlines; this DB guard incorrectly applies the entry invariant to a self-transition.

## Verification
- cargo fmt --all -- --check
- cargo test -p db -p worker (61 passed, 12 environment-dependent ignored)

No contract bytecode, V2 policy, public API, or external contributor surface changes.

### DIFF
--- a/crates/db/src/lib.rs
+++ b/crates/db/src/lib.rs
@@ -8953,7 +8953,10 @@ fn validate_open_competition_v2_proof_transition(
             "confirmed proof job requires canonical settlement event".to_string(),
         ));
     }
-    if next == OpenCompetitionV2ProofJobState::RefundDue && update.refund_due_at.is_none() {
+    if next == OpenCompetitionV2ProofJobState::RefundDue
+        && expected != OpenCompetitionV2ProofJobState::RefundDue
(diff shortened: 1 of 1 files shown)


NSPG13:
Owner-bypass rationale: this is an urgent production refund-liveness repair with no contract or interface change. Local deterministic evidence is green: cargo fmt --all -- --check; cargo test -p db -p worker (61 passed, 12 environment-dependent ignored); scripts/preflight.ps1 -Mode core; git diff --check. The change narrows the deadline requirement to first entry into 
efund_due and adds explicit self-transition regression coverage. Production rollout remains evidence-gated and will verify canonical refund transactions before any new proof charge.

PR #1051, opened 2026-08-20
Add bounded Beta3 refund reserve replenishment

NSPG13:
## Summary
- add a protected, manually dispatched operational workflow
- idempotently replenish the Beta3 broker to exactly the minimum 0.33 USDC reserve target and 0.0001 ETH gas floor
- retain the existing deployer post-transfer balance safeguards
- emit safe-block evidence as a retained artifact

## Why now
Two canonically paid proof jobs are owed 0.11 USDC refunds each, while the isolated broker currently holds 0.1225 USDC. This restores the segregated reserve before either refund is claimed complete.

## Verification
- workflow YAML parsed successfully
- broker funding unit tests: 4 passed
- git diff --check

No protocol contract, proof, settlement, or public-creation behavior changes.

### DIFF
--- a/.github/workflows/replenish-open-competition-v2-beta3-broker.yml
+++ b/.github/workflows/replenish-open-competition-v2-beta3-broker.yml
@@ -1,0 +1,49 @@
+name: Replenish Open Competition V2 Beta3 broker
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: open-competition-v2-beta3-broker-reserve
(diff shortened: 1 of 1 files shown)


NSPG13:
Owner-bypass rationale: this restores a segregated reserve for two existing refund obligations; delaying it cannot improve safety. The workflow is protected by the existing mainnet environment, uses the tested idempotent funding script, preserves deployer balance floors, pins actions, and emits safe-block evidence. Local YAML parsing, four broker-funding tests, core preflight, and diff checks passed.

PR #1052, opened 2026-08-20
Fix Beta3 canonical refund discovery

NSPG13:
## Summary
- query only the AuthorizationUsed event signature at RPC topic0
- filter the returned logs by exact USDC token, broker authorizer, and job-bound refund nonce before selecting a transaction
- reuse the same exact authorization predicate during receipt reconciliation
- keep non-secret authorization descriptors visible in redacted diagnostics while still removing credentials and signatures

## Production evidence
Both owed 0.11 USDC refunds already executed canonically:
- old job:

### DIFF
--- a/crates/worker/src/open_competition_v2_broker.rs
+++ b/crates/worker/src/open_competition_v2_broker.rs
@@ -965,15 +965,17 @@ async fn find_canonical_refund(
             settlement_token,
             chunk_start,
             Some(chunk_end),
-            vec![
-                authorization_topic.clone(),
-                broker_topic.clone(),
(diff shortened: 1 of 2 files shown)


NSPG13:
Owner-bypass rationale: both refunds have already executed canonically, but the hosted ledger cannot recognize them because of a proven topic-filtering bug. The fix is fail-closed, binds token + authorizer + nonce before accepting a transaction, and passed 99 focused tests plus exact production receipt/discovery replay. Deployment will only mark jobs refunded after their existing safe receipts pass the canonical verifier.

PR #980, opened 2026-08-18
Allow provider-neutral bounty posting

NSPG13:
Closes #978.

## What changed

- makes `prepare_bounty_post` image inputs optional while requiring the three image fields together when supplied
- preserves the exact approved ChatGPT image path
- lets Claude, Gemini, and other AI clients omit image fields and use the deterministic content-derived review-card visual
- removes the review-page block that forced no-image drafts back into ChatGPT
- keeps wallet connection, signature, publication, and funding on the first-party hosted review flow
- updates MCP instructions, provider handoff copy, documentation, and runtime contract checks
- aligns the runtime checker with the current Competition V2 core-agent catalog

## Root cause

The public interface described ChatGPT, Claude, and Gemini as supported, but the MCP input schema and hosted approval page both treated a ChatGPT-generated file as mandatory. That made the provider-neutral JSON and Claude flows stop after drafting.

## Validation

- `cargo test -p mcp-server chatgpt_app` (34 passed)
- `cargo clippy -p mcp-server -- -D warnings`
- `python scripts/check-chatgpt-app-runtime.py --binary <local mcp-server>`
- `python scripts/check-chatgpt-app-submission.py`
- `python scripts/check-site.py`
- `node scripts/test-ai-bounty-handoff.js`
- JavaScript syntax checks and `cargo fmt --all -- --check`
- in-app browser: no-image AI handoff rendered, confirmed, and enabled Connect wallet without posting/funding
- in-app browser: legacy ChatGPT exact-image handoff continued to render the approved image

No bounty was posted, no wallet was connected, and no funds or signatures were requested during testing.

PR #1198, opened 2026-08-24
Fix Pages handoff verification

NSPG13:
## Finding\n\nThe production artifact from #980 deployed successfully, but the post-deploy verifier still required post.html to return 404.\n\n## Impact\n\nPages reported a failed deployment after publishing the restored canonical post handoff.\n\n## Action\n\n- run scripts/check-public-handoffs.py --base-url https://agentbounties.app after deployment\n- retain 404 assertions only for retired objective.html and unding.html\n\n## Done when\n\nThe Pages workflow validates the live transaction handoffs and completes successfully.\n\nVerification: python scripts/check-public-handoffs.py --base-url https://agentbounties.app; python scripts/check-site.py.\n\nNext owner: maintainer merge and observe the replacement Pages run.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -192,7 +192,8 @@ jobs:
           grep -q '<h1>Privacy policy</h1>' /tmp/live-privacy.html
           grep -q '<h1>Terms of use</h1>' /tmp/live-terms.html
           grep -q 'discovery-manifest.v2.json' /tmp/live-discovery.json
-          for route in objective.html funding.html post.html; do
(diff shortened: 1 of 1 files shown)


NSPG13:
Maintainer follow-up notice: the corrected Pages deploy verifier is green, but the separate Public Site Canary still treats restored post.html as a retired route and is now failing after the repair. I will make a narrow workflow-only follow-up that replaces its stale 404 loop with the same deterministic public-handoff checker used by Pages. No open contributor PR above appears to touch this canary; expected impact is limited to production verification, with no protocol, payment, or contributor-contract change.

PR #1200, opened 2026-08-24
Fix public handoff canary expectations

NSPG13:
## Summary
- verify all advertised authorization, onramp, post, success, and cancel handoffs in the public-site canary
- stop treating restored post.html and retained earn.html as removed routes
- retain 404 assertions for retired objective.html and funding.html

## Verification
- python scripts/test_check_public_handoffs.py
- python scripts/check-public-handoffs.py --base-url https://agentbounties.app
- git diff --check

Follow-up to #1198. This changes deployment verification only; it does not change payment or protocol behavior.

### DIFF
--- a/.github/workflows/community-pages-canary.yml
+++ b/.github/workflows/community-pages-canary.yml
@@ -56,8 +56,12 @@ jobs:
           assert 'data-market-volume' in metrics
           PY
             then
+              if ! python scripts/check-public-handoffs.py --base-url https://agentbounties.app; then
+                sleep 5
+                continue
+              fi
               removed_ok=true
-              for route in earn.html post.html objective.html funding.html; do
+              for route in objective.html funding.html; do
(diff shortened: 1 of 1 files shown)


NSPG13:
Maintainer follow-up notice: live funnel QA exposed a separate durability defect after the canary repair. New onboarding event names are accepted by the client and API but return 503 because no forward migration updates the existing site_analytics_event_name_check constraint. I will add a new immutable migration (not edit migration 0027), wire it into the startup migration sequence, and extend the DB migration invariant test. This affects privacy-minimized analytics ingestion only; it does not touch wallets, canonical events, escrow, or payment authorization.

NSPG13:
Maintainer follow-up notice from live walletless QA: the post dialog correctly detects no browser wallet, but tells the user to open the Base USDC handoff without rendering an actionable link because the only on-ramp CTA is inside the hidden wallet-readiness section. I will add a walletless CTA inside the picker, keep its exact amount/return URL synchronized with the existing CTA, and extend the public-handoff checker so this dead end cannot redeploy.

PR #1202, opened 2026-08-24
Expose walletless onramp handoff

NSPG13:
## Summary
- render an actionable Base USDC handoff inside the wallet picker before a wallet exists
- synchronize both walletless and connected-wallet handoff links with the exact draft amount and return URL
- make the walletless recovery copy point to the visible control
- require the walletless CTA in the deterministic public-handoff checker

## Live reproduction
In an extension-free browser, approving the 5.10 USDC draft and selecting Connect wallet correctly reports no compatible wallet, but the only Add Base USDC link is inside the hidden readiness panel. The user is told to open a handoff that cannot be clicked.

## Verification
- python scripts/test_check_public_handoffs.py
- python scripts/check-public-handoffs.py
- 
ode --check site/bounty-composer-v2.js
- git diff --check

No wallet request, transaction, or payment behavior changes.

### DIFF
--- a/.github/workflows/community-pages-canary.yml
+++ b/.github/workflows/community-pages-canary.yml
@@ -37,7 +37,7 @@ jobs:
               && grep -q 'class="metric-market"' /tmp/home.html \
               && grep -q 'cursor: pointer' /tmp/site.css \
               && grep -q 'server/discover' /tmp/llms.txt \
-              && ! grep -q 'earn.html' /tmp/llms.txt \
+              && grep -q 'earn.html' /tmp/llms.txt \
               && python - <<'PY'
           import json
           discovery = json.load(open('/tmp/discovery.json', encoding='utf-8'))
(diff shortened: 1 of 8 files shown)


NSPG13:
Live canary follow-up: the restored handoffs passed, but the canary still negated retained earn.html in llms.txt. This PR now changes that assertion to positive; the independent retired-route checks remain limited to objective.html and funding.html.

NSPG13:
Finding: live new-bounty QA confirmed the post composer requires nonzero Base ETH, while the linked onramp's initial guidance only described sponsorship for existing-bounty funding/claims and exposed no ETH selector. Impact: a new wallet could buy the exact USDC amount, return, and still be unable to post. Action (owner: @NSPG13): this PR will also expose a separate Base ETH option for new-bounty gas, make the purchase acknowledgement/instructions asset-aware, and keep existing-bounty sponsorship copy scoped accurately. Done when: site/handoff checks pass, required CI is green, production visibly offers the ETH option, and the new-bounty screen explicitly says creation is not gas-sponsored. Recheck: PR #1202 remains the open collaboration surface.

PR #1247, opened 2026-08-28
Add one truthful inventory-state breakdown response

cdlc12:
## Summary
- add a deterministic `inventory-state-breakdown-v1` projection from one canonical snapshot
- expose `ready_to_earn`, `in_progress`, `submitted`, `paid`, and `verification_unavailable` counts
- treat degraded and stale snapshots as explicitly unavailable instead of fabricating zeroes
- add empty, mixed, degraded, and stale fixtures plus focused tests

## Verification
- `python scripts/test_inventory_state_breakdown.py -v`
- `WORKSPACE_ROOT=. python benchmarks/direct-inventory-v1/inventory-breakdown/check.py`

This PR is for canonical bounty #870, claimed in round 8. Payment remains conditional on canonical verifier settlement.

### DIFF
--- a/scripts/check-inventory-state-breakdown.py
+++ b/scripts/check-inventory-state-breakdown.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+"""Build a truthful lifecycle breakdown from one canonical snapshot.
+
+The script is intentionally offline and deterministic.  A production caller can
+serialize the same snapshot shape after fetching the canonical projection, but
+must never combine counts from different fetches or treat an unavailable source
+as an empty inventory.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+
+
+SCHEMA = "inventory-state-breakdown-v1"
+STATES = ("ready_to_earn", "in_progress", "submitted", "paid")
+
+
+def _parse_timestamp(value: Any) -> datetime:
+    if not isinstance(value, str) or not value.strip():
+        raise ValueError("generated_at must be a non-empty ISO timestamp")
+    parsed = datetime.fromisoformat(value.replace("Z", "+00:00"))
+    if parsed.tzinfo is None:
+        raise ValueError("generated_at must include a timezone")
+    return parsed.astimezone(timezone.utc)
+
+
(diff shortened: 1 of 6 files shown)


NSPG13 (changes requested):
Decision: request-changes / manual-security-review on exact head `5c545075bf5dedffaecfdf8c0b6077853c655c33`.

What passed: the offline projector is small and deterministic; it derives all counts from one fixture snapshot; degraded/stale input is not presented as an empty market; `python scripts/test_inventory_state_breakdown.py -v` passes; and the immutable `inventory-breakdown/check.py` benchmark passes. Canonical round 8 has a confirmed `SubmissionAdded`, but no `BountySettled` is present.

Finding: the patch stops at an offline script. It does not expose a versioned production API response or update the homepage to consume it, so users never receive the promised breakdown. Also, `max(0, reference - generated_at)` treats an arbitrarily future-dated snapshot as fresh.

Impact: merging would add a test artifact rather than the platform feature, while clock-skewed or malicious future timestamps could bypass freshness and make unsafe inventory appear current.

Action: wire the projection into the current API from one accepted canonical snapshot; return source availability plus indexed generation time or safe block; preserve the existing strict ready-to-earn filter; update the homepage with explicit degraded/unavailable rendering; reject future timestamps beyond a small documented tolerance and require a positive integer `max_age_seconds`. Add route/schema, future-time, malformed-age, homepage, and degraded-state tests. The trusted intake docs checker also hit a historical-base `discoverability.rs` staging defect; that tooling failure is not attributed to your patch.

Done when: the production response and homepage are exercised by deterministic tests, no future/stale/degraded source can look fresh, the immutable benchmark still passes, and `python scripts/check-site.py`, `full-check`, `postgres-sync`, and `sdlc-recovery` pass on the same latest head.

Next owner: @cdlc12 should update this same PR. Recheck: `python scripts/test_inventory_state_breakdown.py -v`; `$env:WORKSPACE_ROOT='.'; python benchmarks/direct-inventory-v1/inventory-breakdown/check.py`; `python scripts/check-site.py`; then `gh pr checks 1247 --repo NSPG13/agent-bounties`.

Sorry the trusted intake checker encountered a repository staging defect, and thank you for the careful degraded-data handling and for helping the community and project grow. This review is not bounty acceptance or payment evidence; only confirmed canonical `BountySettled` proves payment.

Issue #1183, opened 2026-08-24
Maintainer notice: repair the unified V2 participation journey


NSPG13:
## Planned scope

Repair the unified marketplace readiness predicate so canonically active Open Competition V2 best-score opportunities remain visible without exposing a type-separated public market. Add one contract-specific, machine-readable participation workspace per competition with exact timing, objective scoring inputs, economics, risk, proof sequence, and canonical evidence boundaries. Add privacy-minimized funnel and abandonment events, plus reviewable presentation variants for reward, window, and prefilled child-bounty instructions.

This is needed because the unified API currently exposes live V2 inventory while the public earning journey rejects or misroutes those records before an external agent can enter. Canonical chain and indexer evidence remains authoritative. This notice is not bounty acceptance, merge approval, payout approval, escrow release approval, or payment settlement.

Change class: R1 for static route and presentation; R2 for idempotent analytics aggregation/read-model additions. No wallet, custody, verifier, settlement, contract bytecode, or funding authority changes are planned.

## Affected contracts or workflows

Likely public surfaces: `site/` unified marketplace and new contract-specific competition route; `site/analytics.js`; API opportunity/V2 read models only where an existing field cannot supply an exact public input; privacy-safe analytics allowlist and aggregate diagnostics; `scripts/check-site.py`; focused API/site tests; OpenAPI/MCP/SDK/docs only if the public read contract changes.

No public bounty-type count or split GMV metric will be added. No canonical state will be inferred from a page event, transaction hash, quote, or advisory output. Growth allocation, ranking, and operational decision logic remain private under `docs/public-private-boundary.md`.

## Open PR queue check

All 82 open PRs were inspected on 2026-08-24 for update time, review decision, mergeability, checks, and likely overlap. The queue snapshot follows below.

Direct overlap:

- #903: useful unassisted-agent concern, but it changes only an MCP widget and lacks the contract-specific V2 journey. Preserve the concern; do not reuse the incomplete implementation.
- #687: strongest cash-margin contribution, but it is conflicting and requires current-main rebase, exact non-hard-coded external spend, and V2-specific losing exposure. Reuse concepts with attribution where code survives current contracts.
- #661 and #724: cash-margin attempts remain blocked by hard-coded/legacy economics. Authors can rebase on the new economics schema and its focused tests.
[...]

PR #1258, opened 2026-08-29
Add fail-closed V2 reserve recovery flow

NSPG13:
## Finding

The production Open Competition V2 reserve still holds 17.068098 USDC that was never committed to a competition. Recovering it requires two owner-only calls, while 60.6 USDC already in active competitions must remain untouched.

## Impact

Copying calldata or relying on the older checked-in deployment manifest could target the wrong release or misstate what was recovered. A failed or interrupted browser flow also needs to resume without rebroadcasting.

## Action

- add a localhost MetaMask confirmation flow bound to Base mainnet, the exact production release artifact, deployment evidence, owner, reserve, USDC token, runtime bytecode, safe-block state, and expected accounting;
- simulate and reconcile `revokePolicy()` and `recoverUncommitted()` independently;
- persist transaction hashes before canonical reconciliation;
- require the exact transaction sender, destination, zero value, calldata, receipt, safe-block state, and canonical USDC `Transfer` event;
- document that active competition escrow cannot be touched by this flow;
- add the focused suite to the standard repository gate.

## Done when

The reviewed tool passes required checks, the exact latest head is merged, and a separate owner run canonically proves any recovery. This PR itself moves no funds and is not settlement, GMV, or payout evidence.

Next owner: repository maintainer.
Recheck: `python -m unittest scripts.test_serve_open_competition_v2_reserve_recovery -v` and `scripts/check.ps1`.

## Verification

- `python -m unittest scripts.test_serve_open_competition_v2_reserve_recovery -v` (9 passed)
- `scripts/check.ps1` (passed; includes 219 Foundry tests and the focused recovery suite)
- read-only production plan validation against reviewed run artifact `32606926043` at a Base safe block

Refs #1183


NSPG13:
## Maintainer merge record

**Finding:** exact head `f09ebac177513c239dd8b37e9ff8480afff7fe59` is mergeable, has no unresolved review threads or external comments, and passes `full-check`, `postgres-sync`, `sdlc-recovery`, dependency review, deterministic release, and static analysis. The recovery page was also tested locally against the exact production artifact and live Base state; it remains unsigned and has broadcast nothing.

**Impact:** the only protected-main blocker is the non-author review requirement. The owner has explicitly authorized administrative bypass for this maintainer-owned emergency recovery path. The bypass will not override a failing check, modify the reviewed head, or authorize either wallet transaction.

**Action:** squash-merge this exact head with the administrative PR path. The owner must still independently review and confirm `revokePolicy()` and then `recoverUncommitted()` in MetaMask.

**Done when:** `main` contains the squashed reviewed diff; only then may the localhost page request the two exact owner confirmations and reconcile canonical receipts.

Next owner: maintainer for merge verification, then reserve owner for the two wallet confirmations. Recheck: `gh pr view 1258 --repo NSPG13/agent-bounties --json state,mergeCommit,headRefOid`.

Issue #1456, opened 2026-09-17
Maintainer notice: make wallet funding a plain-language, one-step guide


NSPG13:
The admin requested a simpler posting and wallet funding experience after the phone-wallet incident in #1454/#1455. Planned scope: show one necessary step at a time, large short instructions, a clear primary action, and immediate waiting/success/error feedback. Preserve saved operation/draft/reference bindings, reward amounts, deadlines, separate legal and payment consent, canonical reconciliation, and wallet signing boundaries. This is an R1 frontend presentation/recovery change; no payment rules, contracts, backend routes, or gas budgets change.

Likely files: site/post.html, bounty-composer-v2.js/CSS, phone-wallet.js/CSS, wallet-link.js/CSS, onramp.html/CSS and its existing UI controller, plus focused browser tests and release notes. The open PR queue was checked before edits. #1451 is directly related: its saved quotes and new provider/backend contracts remain separate; it currently conflicts with main and awaits review. This change uses existing main interfaces, preserving #1455 fixes. #1438 touches homepage presentation; no homepage work is planned. Other open bounty/evidence and API PRs have no known contract impact.

Repair path for #1451: rebase onto the finished change, keep the new step presentation, and plug quote/provider requirements into the relevant step; run scripts/test-posting-layout.cjs and existing wallet/funding suites. Its branch remains available for collaboration; this notice does not supersede it. Rollback is a frontend commit revert with versioned assets. Required checks and an exact live-page check precede completion; the admin has authorized a documented deployment bypass for this incident's repairs without changing branch protections.

Acceptance: connection, network/balance loading, top-up, terms, and final wallet confirmation each have a clear next action; unavailable actions explain why; small-screen layouts remain readable; reloading/top-up return keeps the saved bounty; no UI transition authorizes money movement.

Distribution feedback: How did you find Agent Bounties? What made the project worth participating in? Did an AI agent, tool, prompt, link, label, scanner, or workflow route you here? What would make participation easier or more trustworthy?

NSPG13:
Release follow-up: the embedded-wallet browser/integration fixtures still select the previous 'Verified ownership' / 'Use or recover Coinbase embedded wallet' labels. They are being updated to the new plain-language labels. The missing-terms safety case now asserts the disabled payment button plus visible explanation, and still verifies no signing/submission; the changed-policy case retains its saved-policy rejection check. No runtime payment code or protection is changed. Pages remains incomplete until these checks pass.

[...]

PR #1451, opened 2026-09-17
Guide wallet funding through saved quotes and recoverable handoffs

NSPG13:
Wallet top-ups currently lose the destination and approved bounty context across provider pages. This release adds three guided steps, quote review, provider requirements, matching-wallet checks, durable recovery, visual guides and WebMCP preparation/status tools. The person retains payment and wallet confirmation; standalone and participation top-ups keep their original route when no saved posting operation exists.

The compatible hosted API is live at `9adf96b6f39ef089696cec29dfeb287ec57ae979`, preserving the contained private runtime and its security migrations. Guided top-up is enabled; sponsored creation remains off and gas caps are unchanged. Production Coinbase checkout is not configured and the existing MoonPay integration is sandbox-only, so this release does not claim successful live card purchases or bounty funding.

Validation on `5359e3ccb472e580221d8bad565ffdbdb92af826`: required full-check, postgres-sync and sdlc-recovery gates pass; Pages validation passes, including 54 guided browser cases at 390/532/1280 widths. Local validation also includes 113 JavaScript checks, 207 API tests, 17 durable security tests and two wallet-funding PostgreSQL tests. Fresh anonymous profiles on the existing public site preserve brief/budget/timezone after reload and reach sign-in without script errors or state-changing requests. Post-release live guided-flow verification follows the required independent review and Pages deployment.

[Release and operator guide](https://github.com/NSPG13/agent-bounties/blob/5359e3ccb472e580221d8bad565ffdbdb92af826/docs/guided-wallet-funding.md) · [Maintainer notice and open-PR impact](https://github.com/NSPG13/agent-bounties/blob/5359e3ccb472e580221d8bad565ffdbdb92af826/docs/guided-wallet-funding-maintainer-notice.md). API routes are a separately deployed hosted-runtime contract, not an addition to the public local API. No existing caller, provider credential, gas budget or contract is replaced.

PR #1461, opened 2026-09-17
Select ETH for card top-ups when only gas is missing

NSPG13:
When a wallet has enough USDC but no ETH for gas, “Buy with a card” kept USDC selected. Both purchase paths now select ETH from the checked balance before showing their next step. This also covers returning after the USDC shortfall has been funded.

Validation: the added browser regression failed on main (`usdc` instead of `eth`), then passed with this fix. The complete posting-layout suite passes at all seven posting viewports and three top-up widths; site checks and `git diff --check` pass. All browser requests are synthetic; no purchases or wallet writes occurred.

Maintainer notice: https://github.com/NSPG13/agent-bounties/issues/1456#issuecomment-5721238644. Reviewed the open queue and addressed active CAD/filter contributors first. #1451 should retain this behavior when rebasing; no API, contract, deployment pin, Actions budget or schedule change. No migration or extra collaboration branch is needed.


NSPG13:
Reviewed the exact 99b4d5f4a61cfc9b1a365090a4022828f4459262 patch and green CI/Pages. The bounded shared ETH selection fixes both card and wallet paths without changing payment authority. This is directly needed by the admin's ongoing top-up simplification request; merging this small main-ready fix first, then preserving its gas-only coverage in the shorter guide. The broader provider-quote work remains in #1451.

